Issue #8, opened 2025-01-22
Docs (API, some description of goals, schema/spec drafts)


cmalinmayor:

msschwartz21:
Minimal docs are officially up and running!

Issue #15, opened 2025-01-23
Windows testing bug


msschwartz21:
```
FAILED tests/test_basics.py::test_read_write_consistency[True-edge_attr_dtypes0-node_attr_dtypes0-int8] - AssertionError: 
Arrays are not equal

Mismatched elements: 4 / 5 (80%)
Max absolute difference among violations: 6
Max relative difference among violations: 1.5
 ACTUAL: array([  5,   4, 127,  10,   2], dtype=int8)
 DESIRED: array([ 10,   2, 127,   4,   5], dtype=int8)
```
See https://github.com/funkelab/geff/actions/runs/12935149695/job/36077831912 for more detail

msschwartz21:
All of the spatial graph test are failing on windows, but passing on every other operating system. I took a closer look at the [logs](https://github.com/funkelab/geff/actions/runs/14093744612/job/39476496896) and there are spatial-graph/witty warnings present on windows that aren't present on the other operating systems. 

@tlambert03 this is admittedly a hail mary, but is there any chance you can shed any light on what might be happening here with spatial-graph/witty?

```
_witty_module_905d8d667436483dd448029341226312.cpp

C:\Users\runneradmin\.cython\witty\_witty_module_905d8d667436483dd448029341226312.cpp(36479): warning C4551: function call missing argument list

C:\Users\runneradmin\.cython\witty\_witty_module_905d8d667436483dd448029341226312.cpp(36480): warning C4551: function call missing argument list

C:\Users\runneradmin\.cython\witty\_witty_module_905d8d667436483dd448029341226312.cpp(36481): warning C4551: function call missing argument list

C:\Users\runneradmin\.cython\witty\_witty_module_905d8d667436483dd448029341226312.cpp(36498): warning C4551: function call missing argument list

C:\Users\runneradmin\.cython\witty\_witty_module_905d8d667436483dd448029341226312.cpp(36971): warning C4551: function call missing argument list

C:\hostedtoolcache\windows\Python\3.12.9\x64\Lib\site-packages\spatial_graph\graph\src/graph_lite.h(1337): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data

C:\hostedtoolcache\windows\Python\3.12.9\x64\Lib\site-packages\spatial_graph\graph\src/graph_lite.h(1337): note: the template instantiation context (the oldest one first) is

[...]

PR #20, opened 2025-03-17
Miscellaneous project setup chores

msschwartz21:
Contributes towards #8 by at least getting the documentation framework set up

Currently the docs at https://funkelab.github.io/geff won't work without appending /latest, but that may be fixed once this PR is merged into main?

Testing action needs to be updated to differentiate between networkx and spatial graph tests, but that can probably happen in the networkx PR.


msschwartz21:
Windows testing issue is a known problem #15

PR #21, opened 2025-03-17
Add utility function for validating zarr structure 

msschwartz21:


msschwartz21:
This PR is ready to be merged. We can utilize this function in tests, but I think that can happen in another PR since there are lots of testing updates that should happen.

PR #24, opened 2025-03-17
Initial networkx implementation

cmalinmayor:
Does not include ragged attributes or checking for mixed dtype attributes, or other edge case checking or handling


src/geff/networkx/io.py
```diff
@@ -0,0 +1,159 @@
...
+            the geff  metadata
+    Returns:
+        nx.Graph: The graph that was stored in the geff file format
+    """
+    # open zarr container
+    path = Path(path)
+    assert path.exists(), f"Cannot read graph from {path} because it does not exist"
+    group = zarr.open(path, "r")
+
+    # TODO: validate the metadata using the schema
```
msschwartz21, line 134: Update to use new metadata schema

src/geff/networkx/io.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Args:
+        path (Path): The path to the root of the geff zarr, where the .attrs contains
+            the geff  metadata
+    Returns:
+        nx.Graph: The graph that was stored in the geff file format
+    """
+    # open zarr container
+    path = Path(path)
+    assert path.exists(), f"Cannot read graph from {path} because it does not exist"
```
msschwartz21, line 131: We could use the validation function here to check that the file is ok. Probably under a flag so that it can be disabled.

msschwartz21 (commented):
We're going to leave ragged attributes for the next PR.

PR #54, opened 2025-06-11
Bugfix - restore original position attribute name when loading

cmalinmayor:
# Proposed Change
Currently geff always saves the position attribute in "position." However, it should load it back in with the original name, and it currently does not. This fixes that. I'm also open to keeping the original name in the file system instead of renaming to "position" 

# Types of Changes

- `networkx` implementation

# Checklist


msschwartz21 (approved):
LGTM! I think we should use the original attr name in the filesystem and just store the mapping in the geff metadata. But we could address that change when we make position optional if you want to leave it for later.

Issue #47, opened 2025-05-01
Configure mkdocs: Explicitly list API modules instead of auto-api discovery


cmalinmayor:
And whatever formatting we want. Probably means listing the modules explicitly instead of autoapi'ing the whole src directory.

PR #59, opened 2025-06-13
Expose read_nx and write_nx as top level imports

msschwartz21:
# Proposed Change
- Rename `geff.networkx.io.read` and `geff.networkx.io.write` to `geff.networkx.io.read_nx` and `geff.networkx.io.write_nx`
- Expose both functions as top level imports `from geff import read_nx, write_nx` along with `validate` and `GeffMetadata`
- I deleted the `__init__.py` inside of the networkx submodule since I don't think it's necessary anymore. But if I'm wrong we can add it back in.

# Types of Changes
- New feature or enhancement

- `networkx` implementation

# Checklist

### DIFF
--- a/tests/test_networkx/test_nx_basics.py
+++ b/tests/test_networkx/test_nx_basics.py
@@ -2,7 +2,7 @@ import networkx as nx
 import numpy as np
 import pytest
 
-import geff.networkx.io as geff_nx
+import geff
 
 node_dtypes = ["int8", "uint8", "int16", "uint16"]
 node_attr_dtypes = [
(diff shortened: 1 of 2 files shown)


tests/bench.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 import pytest
 
-import geff.networkx as geff_nx
+import geff.networkx.io as geff_nx
```
cmalinmayor, line 8: It's a little weird to do this instead of just having `import geff` here and `geff.write_nx` below (since we are now recommending that usage pattern)

cmalinmayor: Same comment for the other tests. I'd prefer at least one of them use the new import structure (for both read_nx/write_nx and validate) to make sure it worked.

src/geff/networkx/__init__.py
cmalinmayor, line 1: I suspect we need to update the API docs, since now the networkx read and write functions are not auto-api'ed based on my local preview:
Screenshot 2025-06-22 at 2 45 08 PM
 It's so scuffed anyways that I am in favor of manually enumerating the things that need to be API'ed instead of using the auto-discovery. I think updating everything should be a separate PR (I'll make an issue now), but we should manually add an API page for the networkx functions in this PR.

cmalinmayor (changes requested):
Very close! Just docs and nitpick about using the new imports in tests.

src/geff/networkx/__init__.py
cmalinmayor: #47 Issue already exists :) I renamed to be more explicit

tests/bench.py
msschwartz21: Yup fair point. Pretty sure I just did whatever required the smallest change at the time. Both test files are now update to use a direct import geff.

cmalinmayor (approved):
:troll:

PR #60, opened 2025-06-13
Make position an optional attribute

msschwartz21:
# Proposed Change
- Update schema to make position optional and only validate spatial metadata is position is provided
- Update networkx implementation for make position optional
- Write position attribute with name provided by user, not overwriting to default (I think this is true, but should confirm)

# Types of Changes
- New feature or enhancement

- Specification
- `networkx` implementation

# Checklist


src/geff/networkx/io.py
```diff
@@ -207,8 +212,9 @@ def _set_attribute_values(
             # Get either individual item or list instead of setting with np.array
             val = val.tolist() if val.size > 1 else val.item()
             if nodes:
-                if name == "position" and "position_attr" in graph.graph:
-                    graph_attr = graph.graph["position_attr"]
+                # Do subsequent todos needs to be changed with optional position?
+                if name == "position" and "position_attr" in graph.graph:  # TODO
+                    graph_attr = graph.graph["position_attr"]  # TODO
```
msschwartz21, line 217: @cmalinmayor Does this code block also need to be changed in order to make position optional?

src/geff/networkx/io.py
```diff
@@ -107,14 +108,20 @@ def write(
...
-        for node, data in graph.nodes(data=True):
-            if position_attr not in data:
-                raise ValueError(f"Node {node} does not have position attribute {position_attr}")
+        if position_attr is not None:
+            if position_attr not in node_attrs:
+                raise ValueError(f"Position attribute {position_attr} not found in graph")
+            for node, data in graph.nodes(data=True):
+                if position_attr not in data:
+                    raise ValueError(
+                        f"Node {node} does not have position attribute {position_attr}"
```
cmalinmayor, line 117: If we are allowing the position attribute to be totally missing, we should probably also allow it to be partially missing. Right?

src/geff/networkx/io.py
cmalinmayor: It should be removed, because it is no longer needed. This is the part I added to rename the stored position (always called "position") back to the original name (the `position_attribute`) when we read the graph back in. Since we now save it with the `position_attribute` name, we don't need to handle it differently anymore.

cmalinmayor:
> > I'm not sure about not allowing partially missing positions if we allow fully missing positions.
> 
> By partially missing position, do you mean that some nodes would have positions, but others wouldn't. I think I'm pretty strongly against that. It would be such a pain to handle downstream in other applications. I think the graph can either be spatial or not spatial, but not a mix. Theoretically, if someone really wanted sparse position information they could store it on the graph, but not list it as an official position attribute.

Yes, that is what I meant. I'm fine leaving it strict as long as we keep open mind about it if use cases emerge.

msschwartz21:
@cmalinmayor I had forgotten to actually update the spec in the docs with info about optional position. Can you take a look at that change and confirm the language?

PR #94, opened 2025-07-17
rustworkx backend implementation

JoOkuma:
# Proposed Change
This PR adds `read_rx` and `write_rx` to support a rustworkx backend.

# Types of Changes
- New feature or enhancement

- `rustworkx` implementation

# Checklist

# Further Comments
For the `write_rx` I'm considering using the discussed higher level `dict[str, Array]` writing API.

### DIFF
--- a/src/geff/rustworkx/io.py
+++ b/src/geff/rustworkx/io.py
@@ -1,0 +1,109 @@
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import zarr
+
+import geff
+
+if TYPE_CHECKING:
+    import rustworkx as rx
+
+from geff.metadata_schema import GeffMetadata
+
+
+def write_rx(
+    graph: "rx.PyGraph",
+    path: str | Path,
+    node_id_dict: dict[int, int] | None = None,
+    position_attr: str | None = None,
+    axis_names: list[str] | None = None,
+    axis_units: list[str] | None = None,
+    zarr_format: int = 2,
+    validate: bool = True,
+) -> None:
+    """
+    Write a rustworkx graph to the geff file format
+    """
+
+
+def _set_attribute_values(
+    rx_attrs: list[dict],
+    group: zarr.Group,
+) -> None:
+    """Add attributes in-place to a networkx graph's nodes or edges."""
+    arange = np.arange(len(rx_attrs))
+
+    for name in group.keys():
+        attrs = group[f"{name}/values"][...].tolist()
+
+        try:
+            missing = group[f"{name}/missing"][...]
+            current_indices = arange[~missing]
+        except KeyError:
+            current_indices = arange
+
+        for idx in current_indices.tolist():
+            rx_attrs[idx][name] = attrs[idx]
+
+
+def read_rx(
+    path: Path | str,
+    validate: bool = True,
+) -> tuple["rx.PyGraph", dict[int, int]]:
+    """
+    Read a geff file into a rustworkx graph.
+    Metadata attributes will be stored in the graph attributes
+    and can be accessed via `G.attrs[key]` where G is a rustworkx graph.
+
+    Args:
+        path: Path to the geff file.
+        validate: Whether to validate the geff file.
+
+    Returns:
+        A tuple containing the rustworkx graph and a dictionary mapping node ids to rx ids.
+        The rx ids are 0-indexed.
+    """
+    try:
+        import rustworkx as rx
+
+    except ImportError as e:
+        raise ImportError(
+            "rustworkx is not installed. Please install it with `pip install rustworkx`."
+        ) from e
+
+    path = Path(path)
+
+    if validate:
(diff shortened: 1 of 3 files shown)


msschwartz21:
I need to add this to the PR checklist, but there is a benchmarking script that checks read/write time: https://github.com/live-image-tracking-tools/geff/blob/main/tests/bench.py. New functions will need to be added for new implementations.

JoOkuma:
Hi @msschwartz21 and @cmalinmayor, I consolidated the networkx and rustworkx io into `io_utils.py`.
I added to the benchmark when possible.
It's a bit unfortunate that `rx` speedup is minimal. I profiled, and most of the overhead is in `_ingest_dict_rx` because the way we interact with the dictionaries.

src/geff/rustworkx/io.py
```diff
@@ -0,0 +1,264 @@
...
+    metadata = create_or_update_metadata(
+        metadata,
+        isinstance(graph, rx.PyDiGraph),
+        axes,
+    )
+    metadata.write(group)
+
+
+def _ingest_dict_rx(graph_dict: GraphDict) -> tuple[rx.PyGraph, dict[int, int]]:
+    """Convert a GraphDict to a rustworkx graph."""
```
msschwartz21, line 172: Can you add the returns to the docstring?

src/geff/rustworkx/io.py
```diff
@@ -0,0 +1,264 @@
...
+            for idx, val in zip(current_indices, values, strict=True):
+                edges_data[idx][name] = val
+
+        # Add edges with their properties
+        graph.add_edges_from([(e[0], e[1], d) for e, d in zip(edge_ids, edges_data, strict=True)])
+
+    return graph, to_rx_id_map
+
+
+def read_rx(
```
msschwartz21, line 240: My understanding of the biggest rustworkx sticking point is that difference in how node ids are handled. Can you make sure that the approach is documented somewhere.

tests/bench.py
```diff
@@ -42,11 +43,14 @@ def big_graph_path(tmpdir_factory, big_graph):
     return tmp_path
 
 
-def test_write(benchmark, tmp_path, big_graph):
+@pytest.mark.parametrize(
+    "write_func", [geff_nx.write_nx]
+)  # , geff_rx.write_rx])  # we cannot test rustworkx because the big_graph is a NX graph
```
msschwartz21, line 48: Theoretically, we could write the next graph read it in with rx and then write it out again with rx... All that prep work can be excluded from the benchmarking time

msschwartz21 (changes requested):
I only did a very superficial pass, but generally this is all looking good to me. I asked for a couple additional documentation notes, but I think I'm pretty happy to merge otherwise.

src/geff/rustworkx/io.py
```diff
@@ -0,0 +1,264 @@
...
+        # converting to local rx ids
+        edge_ids = np.vectorize(to_rx_id_map.__getitem__)(graph_dict["edges"])
+        # Prepare edge data with properties
+        edges_data: list[dict[str, Any]] = [{} for _ in edge_ids]
+        indices = np.arange(len(edge_ids))
+
+        for name, prop_dict in graph_dict["edge_props"].items():
+            values = prop_dict["values"]
+            if "missing" in prop_dict:
+                current_indices = indices[~prop_dict["missing"]]
```
cmalinmayor, line 223: Make an issue about doing this for networkx as well to improve efficiency please :)

tests/bench.py
cmalinmayor: This is okay to merge without benchmarking (Morgan and I discussed and agreed)

src/geff/networkx/io.py
```diff
@@ -217,12 +145,11 @@ def _set_property_values(
         ignore = prop_dict["missing"][idx] if sparse else False
         if not ignore:
             # Get either individual item or list instead of setting with np.array
-            val = val.tolist() if val.size > 1 else val.item()
             if nodes:
-                graph.nodes[_id.item()][name] = val
+                graph.nodes[_id.item()][name] = val.tolist()
```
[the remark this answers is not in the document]
cmalinmayor: Both of those other people were me 🙃 and the current me agrees that `.tolist()` everywhere is correct

src/geff/rustworkx/io.py
```diff
@@ -0,0 +1,264 @@
...
+
+    axes = axes_from_lists(
+        axis_names,
+        axis_units=axis_units,
+        axis_types=axis_types,
+        roi_min=roi_min,
+        roi_max=roi_max,
+    )
+
+    metadata = create_or_update_metadata(
```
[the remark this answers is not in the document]
JoOkuma: Indeed, I can't believe this went through. Fixed!

src/geff/rustworkx/io.py
cmalinmayor: 🎊 👍

PR #95, opened 2025-07-17
elaborated error message when provided path doesnt have .zattrs

TeunHuijben:
# Proposed Change
- Added  error message when the provided path doesn't have .zattrs
- Made a little fix in `utils` due to pre-commit

### DIFF
--- a/src/geff/utils.py
+++ b/src/geff/utils.py
@@ -28,17 +28,9 @@ def validate(path: str | Path):
 
     graph = zarr.open(path, mode="r")
 
-    # Check if zattrs are empty or missing
-    if not graph.attrs:
-        raise ValueError(
-            f"No zattrs found in {path}. This may indicate the path is incorrect or "
-            f"zarr group name is not specified (e.g. /dataset.zarr/tracks/ instead of "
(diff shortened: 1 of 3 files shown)


src/geff/utils.py
```diff
@@ -28,6 +28,14 @@ def validate(path: str | Path):
 
     graph = zarr.open(path, mode="r")
 
+    # Check if zattrs are empty or missing
+    if not graph.attrs:
```
cmalinmayor, line 32: The root of a zarr does have attrs, so I don't think this covers the case you wanted to cover. It's still a great check though! I would end this error message at "This may indicate the path is incorrect", and then add another check right below to see if "geff_version" is in the zattrs. If it is not, we can say the second part about the group name not being right.

cmalinmayor (changes requested):

cmalinmayor (approved):
Amazing! I realized I slightly misread the code the first time (my priors were overly influenced by the issue title) but I still think this is better 😁  Thank you!

TeunHuijben:
Thanks! I leave the merging to @cmalinmayor / @msschwartz21

Issue #91, opened 2025-07-17
Create an intermediate function for read/write that implementations can rely on


msschwartz21:
This will reduce code duplication and make it easier for implementations to stay up to date with changes to the spec.

msschwartz21:
Also #129 is contributing towards this goal

melisande-c:
> Also [#129](https://github.com/live-image-tracking-tools/geff/pull/129) is contributing towards this goal

b602b78 adds an intermediate read function.

PR #97, opened 2025-07-17
General `geff` writer helper

JoOkuma:
# Proposed Change

This PR adds a `geff` writer abstract class to simplify the writing of alternative backends.

# Types of Changes

- New feature or enhancement

- `networkx` implementation

# Checklist

- [X] I have read the developer/contributing docs.
- [X] I have written docstrings and checked that they render correctly.

# Further Comments
I had to change more things than I expected.
Partially addresses #91 

cc @msschwartz21


src/geff/writer_helper.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(self, group: zarr.Group, position_attr: str | None):
+        if group.name not in ["/nodes", "/edges"]:
+            raise ValueError(f"Group must be a 'nodes' or 'edges' group. Found {group.name}")
+
+        self._group = group
+        self._position_attr = position_attr
+
+    @abc.abstractmethod
+    def _data_dict_iterator(self) -> Sequence[tuple[Any, dict[str, Any]]]:
```
msschwartz21, line 32: Not necessary since we can just iterate over self._data

src/geff/writer_helper.py
```diff
@@ -0,0 +1,104 @@
...
+
+class BaseWriterHelper(abc.ABC):
+    """
+    Base class for writing geff attributes to a zarr group.
+
+    Important:
+        This function does not create the `ids` attribute.
+
+    Args:
+        group: The zarr group to write the attribute to.
```
msschwartz21, line 17: Just checking my understanding and then maybe we need to add more documentation. This is assumed to be an already opened zarr group which is either nodes or edges inside of the root geff group?

src/geff/networkx/io.py
```diff
@@ -40,30 +41,39 @@ def get_roi(graph: nx.Graph, position_attr: str) -> tuple[tuple[float, ...], tup
...
+        group: zarr.Group,
+        data_sequence: Sequence[tuple[Any, dict[str, Any]]],
+        position_attr: str | None = None,
+    ):
+        super().__init__(group, position_attr)
+        self._data = data_sequence
+        ids = [idx for idx, _ in data_sequence]
+        if len(ids) > 0:
+            self._group["ids"] = np.asarray(ids)
+        elif self._group.name == "/nodes":
```
msschwartz21, line 65: This logic is just to handle if there are no nodes or no edges? Can you add a comment?

src/geff/networkx/io.py
```diff
@@ -40,30 +41,39 @@ def get_roi(graph: nx.Graph, position_attr: str) -> tuple[tuple[float, ...], tup
...
     """
-    return list({k for n in graph.nodes for k in graph.nodes[n]})
-
 
-def get_edge_attrs(graph: nx.Graph) -> list[str]:
-    """Get the attribute keys present on any edge in the networkx graph. Does not imply
-    that the attributes are present on all edges.
+    def __init__(
+        self,
+        group: zarr.Group,
```
tlambert03, line 56: one thing to consider (for slightly increased generality) would be to not take the group as an input to the BaseWriterHelper ... but rather have the consumer of the writer pass it as an argument to write_attrs?

so the class just worries about how to gather the right information... but not where it's going

src/geff/writer_helper.py
```diff
@@ -0,0 +1,104 @@
...
+        Must match the index ordering
+        """
+
+    @abc.abstractmethod
+    def _attr_names(self) -> list[str]:
+        """
+        Methods to iterate over the attribute names to be written.
+        """
+
+    def write_attrs(self) -> None:
```
msschwartz21, line 44: So this being generalizable seems to rely on the ability of the in memory representation being able to provide the Sequence data structure of tuples of (id, attr_dict). I just want to double check that there aren't any edge cases that we might be missing where that would be inconvenient. Although if I think of working with a dataframe in memory this would still work.

msschwartz21 (commented):
I'm going to start with a partial review just to give feedback on a few things. Generally a couple more comments explaining the logic of the approach in a few places would be helpful :pray: 

I'm glad to see that the swap worked right off the bat with networkx. However since we're going to reuse this I would like tests for the base class. 

The API of the class feels a little funny to me because I would normally expect the `write_attrs` call to take the actual data not the initialization of the class. But this is a personal preference/style thing so I'm open to other opinions.

src/geff/networkx/io.py
msschwartz21: I agree. Let's put the destination in write attrs

src/geff/networkx/io.py
JoOkuma: I added a comment:

```python
        # special corner cases where the graph is empty the ids must still have the right dimension (N,) for nodes, (N, 2) for edges
        elif group.name == "/nodes":
            group["ids"] = np.empty((0,), dtype=np.int64)
        elif group.name == "/edges":
            group["ids"] = np.empty((0, 2), dtype=np.int64)
        else:
 ```

src/geff/writer_helper.py
JoOkuma: An iterator would still work as long as you could iterate multiple times through it.

src/geff/networkx/io.py
JoOkuma: I moved, and it got much simpler. I don't even need to create a new class for networkx.

msschwartz21 (approved):
🎉

PR #98, opened 2025-07-17
Test: Fixture for path to example GEFF file

melisande-c:
# Proposed Change
Add pytest fixture for a path to example data and the expected graph attributes. This can be used as an initial first pass test for new backends, an example test for the networkx API has been added.

# Types of Changes
- Tests

# Checklist


tests/test_networkx/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+        "extra_node_attrs": {},
+        "edge_attrs": {"score": scores, "color": colors},
+        "directed": directed,
+        "axis_names": axis_names,
+        "axis_units": axis_units,
+    }
+
+
+@pytest.fixture()
+def path_w_expected_graph_attrs(tmp_path) -> tuple[Path, GraphAttrs]:
```
msschwartz21, line 80: I'm wondering if it would be better to just have this as a normal function so that we can take advantage of the parametrization of different dtypes?

tests/test_networkx/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+
+    directed = True
+    graph_attrs = create_dummy_graph_attrs(
+        node_dtype="int8",
+        node_attr_dtypes={"position": "double"},
+        edge_attr_dtypes={"score": "float64", "color": "uint8"},
+        directed=directed,
+    )
+
+    # TODO: write zarr from scratch to remove test dependency on networkx backend?
```
msschwartz21, line 99: I'm ok with doing this through networkx because I suspect we will always ship geff minimally with networkx

msschwartz21 (changes requested):
This looks great! Let's move conftest.py into the tests/ root so that it's available to the other backends. Otherwise I think we should fully refactor the networkx tests to use this new framework 🎉

melisande-c:
> Let's move conftest.py into the tests/ root so that it's available to the other backends. 

Oops! I have fixed.

I have added the requested changes. The fixture has been updated to be a function to make it parameterisable, it is still a fixture so that it can automatically receive the `tmp_path` fixture. I refactored the original networkx tests to use this new fixture, and removed the backend template function. I will add an issue to add documentation for how new backends can use this for their own tests!

msschwartz21 (approved):

Issue #83, opened 2025-07-17
Should time be a special attribute?


msschwartz21:
Should time be a special attribute like position? What special features would be needed? Or should it be baked into position?

cmalinmayor:
I think I will draft an addendum to the spec saying that time is a special attribute like position @msschwartz21, I think we need this sooner rather than later. Sound good?

msschwartz21:
- Axis type (ala ome metadata) should be either time or spatial and is optional. Including channel is still tbd 
- Axes are separate groups in zarr
- Document that we are mimicking ome zarr metadata so if someone has raw or segmentation data they can use ome zarr metadata to keep the axes in sync between graph and image data #86 
- [ ]  Should scale or coordinate transform be included in the spec

PR #99, opened 2025-07-17
Replace axis_names and axis_units with OME NGFF axes

mkitti:
- **Prettify geff-schema.json for future edits**
- **Use OME-NGFF axes instead of axis_{names**
- **Axes: Change min to 1**

# Proposed Change
Instead of uses the `axis_names` and `axis_unit` keys, use a slightly modified `axes` from OME-NGFF 0.5
https://ngff.openmicroscopy.org/0.5/index.html#axes-md

# Types of Changes
Changes the specification

- Specification

# Checklist

# Further Comments


msschwartz21 (commented):
We autogenerate the schema using a pydantic BaseModel so to make this change we would need to update the model: https://github.com/live-image-tracking-tools/geff/blob/main/src/geff/metadata_schema.py. Can you take a look at what would need to be changed there?

mkitti:
Maybe we should just depend on an existing OME pydantic model, let me see what I can find.

One example would be pydantic-ome-ngff:
https://github.com/JaneliaSciComp/pydantic-ome-ngff/blob/06dc31cfb734d278cd52027d1b9e633edc0e67b7/src/pydantic_ome_ngff/v04/axis.py#L125

Another one would be ome-zarr-models-py:
https://github.com/ome-zarr-models/ome-zarr-models-py/blob/main/src/ome_zarr_models/common/axes.py

msschwartz21:
This PR has been replaced by #121 and the metadata restructure forthcoming in #83

PR #100, opened 2025-07-17
Prettify geff-schema.json for future edits

mkitti:
This just runs the specification through `jq` to pretiffy it so that future changes produce easy to read diffs


msschwartz21:
That is much easier to read! We autogenerate the schema whenever we make changes to the spec so I think we would need to update scripts/export_json_schema.py to maintain this format.

mkitti:
I modified scripts/export_json_schema.py so that it now it exactly produces the expanded geff-schema.json

Issue #78, opened 2025-07-17
failing test with zarr 3.1


tlambert03:
on a new install, got zarr v3.1:

```
_______________________________________ TestMetadataModel.test_read_write ________________________________________

self = <test_metadata_schema.TestMetadataModel object at 0x106ab4270>
tmp_path = PosixPath('/private/var/folders/30/fpzwkh4103s6xj67y2k0q4pr0000gn/T/pytest-of-talley/pytest-1/test_read_write0')

    def test_read_write(self, tmp_path):
        meta = GeffMetadata(
            geff_version="0.0.1",
            directed=True,
            position_attr="position",
            roi_min=[0, 0, 0],
            roi_max=[100, 100, 100],
            axis_names=["t", "y", "x"],
            axis_units=["min", "nm", "nm"],
            extra=True,
        )
        zpath = tmp_path / "test.zarr"
>       group = zarr.open(zpath, "a")
                ^^^^^^^^^^^^^^^^^^^^^
E       TypeError: open() takes from 0 to 1 positional arguments but 2 were given
```

downgrading to v3.0 fixes

PR #115, opened 2025-07-18
refactor: use dependency groups instead of extras for dev dependencies

tlambert03:
This PR does a couple things

1. uses [PEP 735 Dependency groups](https://peps.python.org/pep-0735/) in pyproject.toml (instead of `project.optional-dependencies`).  This is probably semantically more accurate (since these are dev dependencies that your end users will likely never install via extras).  It also makes it simpler to just `uv sync`.    Pip also supports groups now as well with `pip install -e . --group dev`.  (I also add `uv.lock` to gitignore... so users won't have to mind it)
1. updates CI to use groups instead of extras
1. Sets warnings to errors in pytest.ini options.  (Would have caught #78, because v3.0 emitted a warning about not using mode).

(I also ran pre-commit and merged in changes)


msschwartz21:
Quick clarifying question, if we wanted to add optional dependencies like rustworkx for that implementation, we would still use project.optional-dependencies in that case right?

PR #129, opened 2025-07-18
feat: FileReader class to read zarr to intermediate dict format

melisande-c:
# Proposed Change
Adding a file reader class that allows a subset of the zarr to be read. Follows a builder pattern, this allows each `node_prop` and `edge_prop` to be read when required. When the `build` method is called, output boolean masks can be provided to read only a subset of nodes and edges.

The `build` method outputs a new intermediate dictionary representation of the graph defined as a `TypedDict`, this mirrors the structure in the zarr file. This allows for reusability of the `FileReader` for different backends.

A `read_to_dict` function has been added that outputs the intermediated dict representation. The `read_nx` function now uses this and can also accept a list of `node_props` and `edge_props` so that not all the properties need to be read.

Tests have been added for the `FileReader`, it is also tested through the `read_nx` texts.

# Types of Changes
- New feature or enhancement

- `networkx` implementation

# Checklist

# Further Comments

### DIFF
--- a/src/geff/geff_reader.py
+++ b/src/geff/geff_reader.py
@@ -135,15 +135,17 @@ class GeffReader:
         Returns:
             GraphDict: A graph represented in graph dict format.
         """
-        nodes = np.array(self.nodes[node_mask if node_mask is not None else ...])
+        nodes = np.array(self.nodes[node_mask.tolist() if node_mask is not None else ...])
         node_props: dict[str, PropDictNpArray] = {}
         for name, props in self.node_props.items():
             node_props[name] = {
-                "values": np.array(props["values"][node_mask if node_mask is not None else ...])
+                "values": np.array(
+                    props["values"][node_mask.tolist() if node_mask is not None else ...]
+                )
             }
             if "missing" in props:
                 node_props[name]["missing"] = np.array(
-                    props["missing"][node_mask if node_mask is not None else ...],
+                    props["missing"][node_mask.tolist() if node_mask is not None else ...],
                     dtype=bool,
                 )
 
@@ -160,11 +162,13 @@ class GeffReader:
         edge_props: dict[str, PropDictNpArray] = {}
         for name, props in self.edge_props.items():
             edge_props[name] = {
-                "values": np.array(props["values"][edge_mask if edge_mask is not None else ...])
+                "values": np.array(
+                    props["values"][edge_mask.tolist() if edge_mask is not None else ...]
+                )
             }
             if "missing" in props:
                 edge_props[name]["missing"] = np.array(
-                    props["missing"][edge_mask if edge_mask is not None else ...],
+                    props["missing"][edge_mask.tolist() if edge_mask is not None else ...],
                     dtype=bool,
                 )
 


melisande-c:
### TODO

- [ ] If nodes are masked then edges on those nodes have to be masked. 
- [ ] An example of reading a masked version of a graph needs to be added.

msschwartz21 (changes requested):
I don't know what the right vocabulary is, but we should synchronize between this PR and @cmalinmayor writer helper PR to have standard names for the dict of arrays format and the sequence of nodes format.

Also this is a preliminary review, but I have a comment that may change the API so I want to get that out first.

The goal here is that we can load props one at a time and cache them as we go before returning a graph with all the props. We expect that if someone wants the graph with all properties they will either use the FileReader and call read_x_prop for each property or they will use the higher level function read_to_dict which takes care of that for them. Correct?

I think there is an entrypoint that is missing. For example, I want to filter my graph based on t so I start by initializing a file_reader, getting t and then constructing the boolean mask for the nodes that I want. Then I want to get the graph will all properties using the bool mask as a filter. Currently to do that, I would also need to go and load each of the other properties manually. So maybe we need a different version of build or a different higher level method on file_reader that will also handle loading props that might not be loaded yet.

src/geff/file_reader.py
```diff
@@ -0,0 +1,208 @@
...
+class FileReader:
+    """
+    File reader class that allows subset reading to an intermediate dict representation.
+
+    The subsets can be a subset of node and edge properties, and a subset of nodes and
+    edges.
+
+    Example:
+        >>> from pathlib import Path
+        ... from geff.read_file import FileReader
```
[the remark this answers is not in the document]
melisande-c: oops yes I changed the file name thanks

src/geff/file_reader.py
```diff
@@ -56,44 +57,59 @@ def __init__(self, path: Path, validate: bool = True):
...
         if "props" in self.group["edges"].keys():
             edge_props_group = zarr.open_group(self.group.store, path="edges/props", mode="r")
             self.edge_prop_names: list[str] = [*edge_props_group.group_keys()]
         else:
             self.edge_prop_names = []
 
-    def read_node_props(self, name: str):
+    def read_node_props(self, names: list[str] | None = None):
         """
         Read the node property with the name `name` from a GEFF.
```
msschwartz21, line 76:
```suggestion
        Read the node property with the name `name` from a GEFF.
        
        If no names are specified, then all properties will be loaded
```

src/geff/file_reader.py
```diff
@@ -56,44 +57,59 @@ def __init__(self, path: Path, validate: bool = True):
...
+            prop_group = zarr.open_group(self.group.store, path=f"nodes/props/{name}", mode="r")
+            prop_dict: PropDictZArray = {"values": prop_group["values"]}
+            if "missing" in prop_group.keys():
+                prop_dict["missing"] = prop_group["missing"]
+            self.node_props[name] = prop_dict
 
-    def read_edge_props(self, name: str):
+    def read_edge_props(self, names: list[str] | None = None):
         """
         Read the edge property with the name `name` from a GEFF.
```
msschwartz21, line 96:
```suggestion
        Read the edge property with the name `name` from a GEFF.
        
        If no names are specified, then all properties will be loaded
```

msschwartz21 (commented):
I like this better! Filtering will still take some additional thought, but I think it's more accessible now. Let's have @cmalinmayor take a look next.

src/geff/file_reader.py
```diff
@@ -0,0 +1,217 @@
...
+import numpy as np
+import zarr
+from numpy.typing import NDArray
+
+from . import utils
+from .dict_representation import GraphDict, PropDictNpArray, PropDictZArray
+from .metadata_schema import GeffMetadata
+
+
+class FileReader:
```
cmalinmayor, line 12: Total nitpick but I would prefer "GeffReader." FileReader is very generic. And I think changing it is now or never 😆

src/geff/file_reader.py
```diff
@@ -0,0 +1,217 @@
...
+            node_mask (np.ndarray of bool): A boolean numpy array to mask build a graph
+            of a subset of nodes, where `node_mask` is equal to True. It must be a 1D
+            array of length number of nodes.
+            edge_mask (np.ndarray of bool): A boolean numpy array to mask build a graph
+            of a subset of edge, where `edge_mask` is equal to True. It must be a 1D
+            array of length number of edges.
+        Returns:
+            GraphDict: A graph represented in graph dict format.
+        """
+        nodes = np.array(self.nodes[node_mask.tolist() if node_mask is not None else ...])
```
cmalinmayor, line 134: I  believe you can do this in pure numpy with: `self.nodes[node_mask != 0]` (and then `props["values"][node_mask != 0]`, etc.)

src/geff/networkx/io.py
```diff
@@ -150,7 +157,30 @@ def _set_property_values(
                 graph.edges[source, target][name] = val
 
 
-def read_nx(path: Path | str, validate: bool = True) -> nx.Graph:
+def _ingest_dict_nx(graph_dict: GraphDict):
+    metadata = graph_dict["metadata"]
+
+    graph = nx.DiGraph() if metadata.directed else nx.Graph()
+    for key, val in metadata:
+        graph.graph[key] = val
```
cmalinmayor, line 165: This will conflict with the latest main, where we have removed storing attributes on the graph and instead just return the metadata with the graph

cmalinmayor (changes requested):
Very close I agree! If I'm right about the numpy mask indexing, it will be much more efficient than python list comprehension, so I would definitely do that. With that update and resolving conflicts with main I think it'll be ready!

src/geff/file_reader.py
melisande-c: `self.nodes` and `props["values"]` are actually currently a `zarr.Array` for some reason I thought it didn't accept bool arrays for fancy indexing, but I've realised I don't need the `.tolist`. For testing if it's `None` or not I don't know what you mean, but indexing a zarr with a boolean array that is all true is much slower than the `Ellipses` `...`.

cmalinmayor: Oooh I think I misread the code my bad. There's more going on in that line than I thought! Yes, totally agree now that with removing the tolist we are at the best we can do

cmalinmayor (approved):
@melisande-c I removed the `tolist()` calls, and turns out they were necessary on some OSes and version combinations, so I added them back in. I also renamed the reader to `GeffReader`. I'm happy to merge now, and if the efficiency of the masking becomes a problem we can address it later. I won't be in at the very beginning of the day tomorrow, but I'll approve this and you/Morgan can merge when you are comfortable :)
Codecov is going crazy again (I actually went to run locally and went down a rabbit hole documented in #170) but your tests look thorough.

PR #137, opened 2025-07-18
Add a testing utility function for comparing two geffs on disk

msschwartz21:
# Proposed Change
This function checks that the types and shapes of the zarr arrays contained within a geff match between two different geffs.

Also updates the networkx read/write consistency test to use this function. When we update the rest of the networkx tests to use fixtures, we should use this function there as well.

# Types of Changes
- Tests

# Checklist


msschwartz21:
This PR has uncovered some issues with the current networkx x implementation. Beginning in #139 ids that are not a string are all cast to int64. This new test function asserts that the dtype should be the same so it fails on the cases where ids are an int dtype. Additionally, I think the same thing is happening to props/values. 

Do we care if something that is int16 becomes int64 or are we happy as long as int-like dtype is maintained?

msschwartz21:
@DragaDoncila this is the PR with the new test function. Lmk if you think it should live somewhere else for maximal utility

CONTRIBUTING
```diff
@@ -14,6 +14,9 @@ To run tests
 pixi run test
 ```
 
+### Conventions
+We provide fixtures for example geff files that are exposed in conftest.py. If you use these fixtures, you can use `tests.utils.check_equiv_geff` to compare the geff generated by the fixture with the geff generated by your implementation.
```
DragaDoncila, line 18: `conftest.py` doesn't actually exist at the moment, so we could mention it for future-proofing purposes or just refer to the `_pytest_plugin` file.

```suggestion
We provide fixtures for example geff files that are exposed in `src/geff/_pytest_plugin.py` or `tests/conftest.py`. If you use these fixtures, you can use `tests.utils.check_equiv_geff` to compare the geff generated by the fixture with the geff generated by your implementation.
```

DragaDoncila:
@msschwartz21 I think it does make sense to expose this one in the pytest plugin

msschwartz21:
Tests are still failing but I think it may get fixed by #148 or at least I will wait for that PR to be merged before investigating further.

src/geff/utils.py
```diff
@@ -89,3 +89,38 @@ def validate(path: str | Path):
...
+        # Check that properties in each geff are the same
+        assert set(ga["props"]) == set(gb["props"])
+
+        # Check shape and dtype of each prop
+        for prop in ga["props"]:
+            if "missing" in ga[f"props/{prop}"]:
+                assert ga[f"props/{prop}/missing"].shape == gb[f"props/{prop}/missing"].shape
+                assert ga[f"props/{prop}/missing"].dtype == gb[f"props/{prop}/missing"].dtype
+            assert ga[f"props/{prop}/values"].shape == gb[f"props/{prop}/values"].shape
+            assert ga[f"props/{prop}/values"].dtype == gb[f"props/{prop}/values"].dtype
```
georgeoshardo, line 126: Apologies if this is already addressed somewhere.  Won't this fail if there is no props on either of the graph_groups?

e.g `check_equiv_geff('mouse-20250719.zarr/tracks', 'mouse-20250719.zarr/tracks')` won't work. 

I've done:

```suggestion
        ga_has_props = "props" in ga
        gb_has_props = "props" in gb

        assert ga_has_props == gb_has_props

        if ga_has_props:
            # Check that properties in each geff are the same
            assert set(ga["props"]) == set(gb["props"])

            # Check shape and dtype of each prop
            # Check shape and dtype of each prop
            for prop in ga["props"]:
                if "missing" in ga[f"props/{prop}"]:
                    assert ga[f"props/{prop}/missing"].shape == gb[f"props/{prop}/missing"].shape
                    assert ga[f"props/{prop}/missing"].dtype == gb[f"props/{prop}/missing"].dtype
                assert ga[f"props/{prop}/values"].shape == gb[f"props/{prop}/values"].shape
                assert ga[f"props/{prop}/values"].dtype == gb[f"props/{prop}/values"].dtype
```

which I think is enough?

cmalinmayor: Oh yes good catch! Thanks :)

CONTRIBUTING
cmalinmayor: I suspect this needs to be further updated after today's decisions about the locations and exposing of tests/validation functions

georgeoshardo:
What is the expected behaviour when two graphs are identical, except for one having a 'missing' group full of only False?

Consider:

```python
g, meta = read_nx(mouse-20250719.zarr) # This has a missing group full of False in nodes/props/track_id
write_nx(g, "savetest.geff", metadata=meta, zarr_format=2)
check_equiv_geff(geff_path, "savetest1.geff")
```

This will fail when trying to `assert ga[f"props/{prop}/missing"].shape == gb[f"props/{prop}/missing"].shape` 

Either the two geffs should be considered the same, or the nx round trip should preserve redundant `missing` information.

msschwartz21:
Ok so tests are still failing for networkx on node ids dtype. I think this is what is happening:
- We start with a graph that we have written with node ids with type int16
- Upon reading those ids into networkx they become normal ints 
- When we write that graph, the list of int ids ends up int64 which I'm guessing is numpy default behavior

We could instead soften the assertion to intlike or floatlike. Does that still seem helpful?

georgeoshardo:
> We could instead soften the assertion to intlike or floatlike. Does that still seem helpful?

In napari-geff I am also giving up on ensuring perfect type symmetry during reading and writing. So I am just asserting that the '`kind`s are the same and that the `itemsize` of the dtype is at least higher than before.

msschwartz21:
@tlambert03 can you take a look at the pylance errors after merging? It's all basically the same thing, but I'm not sure how to fix it.

tlambert03:
it's both pylance and mypy... and the issue is the same thing as https://github.com/live-image-tracking-tools/geff/pull/267#discussion_r2257880646; namely: indexing into a zarr group with `__getitem__` has the possibility to return *either* a zarr Group, or a zarr Array.  So, you will get type errors if you simply use `zarr.open` or `obj.__getitem__` and then use `Array`-specific or `Group`-specific APIs on the object.

either use `zarr.open_group(...)`, or `zarr.open_array(group, path=...)`, or check isinstance, or, if you've previously validated the object, feel free to just `cast()`

msschwartz21:
Correctly testing check_equiv_geff will require closing https://github.com/live-image-tracking-tools/geff/issues/299 so that we can write an in memory geff back to disk

src/geff/testing/_utils.py
```diff
@@ -58,3 +66,69 @@ def nx_is_equal(g1: nx.Graph, g2: nx.Graph) -> bool:
...
     else:
         return False
+
+
+def check_equiv_geff(store_a: StoreLike, store_b: StoreLike) -> None:
+    """This function compares two geffs, typically a starting fixture geff with
+    the output of an implementation.
+
+    This tests focuses on maintaining shape and dtype consistency. It does not
+    assert element wise equality. path_a is assumed to be the "correct" geff.
```
cmalinmayor, line 76:
```suggestion
    assert element wise equality. store_a is assumed to be the "correct" geff.
```

src/geff/testing/_utils.py
```diff
@@ -58,3 +66,69 @@ def nx_is_equal(g1: nx.Graph, g2: nx.Graph) -> bool:
...
+    # Run validation first so that we don't hit issues with basic structure
+    validate_structure(store_a)
+    validate_structure(store_b)
+
+    for graph_group in [_path.NODES, _path.EDGES]:
+        ga = expect_group(open_storelike(store_a), graph_group)
+        gb = expect_group(open_storelike(store_b), graph_group)
+
+        # Check ids
+        ids_a = np.asarray(ga[_path.IDS])
```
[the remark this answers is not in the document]
msschwartz21: If we don't cast to array, ga[_path.IDS].shape causes a type error because it could be either a group or an array. I'll see if casting to zarr.Array without loading into memory can also fix the problem.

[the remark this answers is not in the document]
cmalinmayor: Yeah, loading all the data into memory just to make typing happy seems like a bad idea. This is what the ensure_array and ensure_group helpers are for?

cmalinmayor:
Numpy bool strikes again 🙃

tests/test_networkx/test_nx_basics.py
```diff
@@ -62,6 +64,18 @@ def test_read_write_consistency(
...
+    out_zarr = tmp_path / "write-out.zarr/tracks"
+    geff.write_nx(
+        graph,
+        out_zarr,
+        axis_names=list(graph_props["axis_names"]),
+        axis_units=list(graph_props["axis_units"]),
+    )
+
+    # Check consistency with source fixture
+    # check_equiv_geff(path, out_zarr)
```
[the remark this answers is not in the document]
msschwartz21: I'm just removing this and waiting until we update all the tests to use this function when appropriate.

src/geff/testing/_utils.py
```diff
@@ -58,3 +65,70 @@ def nx_is_equal(g1: nx.Graph, g2: nx.Graph) -> bool:
...
+        b_props = set(expect_group(gb, _path.PROPS))
+        if a_props != b_props:
+            raise ValueError(
+                f"{graph_group} properties: a ({a_props}) does not match b ({b_props})"
+            )
+
+        # Check shape and dtype of each prop
+        for prop in a_props:
+            a_missing = "missing" in expect_group(ga, f"{_path.PROPS}/{prop}")
+            b_missing = "missing" in expect_group(gb, f"{_path.PROPS}/{prop}")
```
tlambert03, line 110: not a suggest for this PR, but just a thought... 

I can see a lot of times where we are re-duplicating things across the library.  (i.e. after validation has been run, re-running expect etc...).  I wonder whether it might be useful to have a `GeffFile` class (something like `TiffFile` or `ND2File`) with dedicated methods (like `nodes() -> zarr.Group` and `edges() -> zarr.Group` and `node_ids() -> zarr.Array`).  These would be both "convenient" (in the sense that you don't need to use strings or imports from `_path` to access canonical parts of a geff file) and also useful for typing, since you can validate once and cast everywhere else.  

I'll play with it...

PR #141, opened 2025-07-18
Read/write ignored attributes to avoid losing them

### LINKED ISSUE #110: Don't lose other metadata in zattrs when reading/writing geffs
We're not mandating anything or using the extra metadata in zattrs but we should make sure if someone loads then save a geff that they don't lose that metadata.

ilan-theodoro:
# Proposed Change
`geff` allows extra attributes inside `zattrs`. However, they are simply ignored. This PR solves it by storing the extra attributes. Closes #110.

# Types of Changes
- Tests

- Specification

# Checklist


cmalinmayor (approved):
Thank you!! 🎉

PR #154, opened 2025-07-19
Add current geff version as default value in GeffMetadata

lxenard:
# Proposed Change
As the title says, when not provided, geff version in GeffMetadata defaults to the current version of geff
Solves #107 .

# Types of Changes
- New feature or enhancement

- Specification

# Checklist

- [x] I have updated the GeffMetadata and the json schema using `pixi run update-schema` if necessary.

# Further Comments


msschwartz21 (changes requested):
My only concern with this PR is that "geff_version" is no longer listed as required in the schema (based on looking at geff-schema.json`). I suspect there just needs to be a change to the pydantic options for Field. Can you look into it and see if we can both have a default and make it required?

lxenard:
I "hacked" the issue by setting the default value of geff_version in an init in GeffMetadata. It behaves as expected: the field stays required in the doc, no validation error are raised at GeffMetadata creation with or without the field, and the correct version is set.
I still think @tlambert03 should check that what I did is ok because I'm clearly out of my comfort zone, and I have no clue if this hack could have nasty side effects.

tlambert03:
The main downside of this approach is that you loose the "free" signature of the `__init__` method that would would have otherwise gotten.  So IDEs will no longer be able to inspect/autocomplete the arguments when you're creating a class (and type checkers can't check it).  The only way I know to customize init without losing all of that is to copy the full typing of every field into the custom init , which is a bummer.


I can take a look at this all tomorrow

lxenard:
Let me know when you have 5min to discuss this, I have a new proposal

PR #155, opened 2025-07-19
Add display_hints to geff schema

funkey:
# Proposed Change
Add `display_hints` to `geff` metadata.

# Types of Changes
- New feature or enhancement

- Specification

# Checklist

- [x] I have updated the GeffMetadata and the json schema using `pixi run update-schema` if necessary.


msschwartz21 (commented):
This seems reasonable to me although the people who were more invested in visualization/tool compatibility should also take a look.

My only request is that we should validate that the properties specified in the DisplayHint fields are also present in the axes metadata. Technically this means the user is forced to declare them as spatiotemporal and not just use any property, but I think that is a reasonable constraint to avoid pointing to properties for display that don't exist.

cmalinmayor (approved):
Looks great, thanks @funkey

Issue #157, opened 2025-07-19
GEFF 0.3.0 breaks `pytest`


yfukai:
Updating `geff` to `0.3.0` causes an error in `pytest` importing the package:
I confirmed this is not the case with `0.2.1`.

```
File "/Users/fukai/projects/laptrack/.venv/lib/python3.13/site-packages/geff/__init__.py", line 8, in <module>
    from .metadata_schema import GeffMetadata
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "/Users/fukai/projects/laptrack/.venv/lib/python3.13/site-packages/_pytest/assertion/rewrite.py", line 170, in exec_module
    exec(co, module.__dict__)
    ~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fukai/projects/laptrack/.venv/lib/python3.13/site-packages/geff/metadata_schema.py", line 26, in <module>
    with (files(geff) / "supported_versions.yml").open() as f:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/importlib/resources/_adapters.py", line 139, in open
    raise FileNotFoundError("Can't open orphan path")
FileNotFoundError: Can't open orphan path
```

PR #158, opened 2025-07-19
Remove supported versions list to solve "File not found" error in pytest

yfukai:
# Proposed Change
Solves #157 by removing `SUPPORTED_VERSIONS` from `metadata_schema.py` and adding regex pattern that generally matches with a version string.

# Types of Changes

# Checklist

### DIFF
--- a/tests/test_agnostic/test_metadata_schema.py
+++ b/tests/test_agnostic/test_metadata_schema.py
@@ -1,11 +1,35 @@
+import re
+
 import pydantic
 import pytest
 import zarr
 
-from geff.metadata_schema import Axis, GeffMetadata, write_metadata_schema
+from geff.metadata_schema import VERSION_PATTERN, Axis, GeffMetadata, write_metadata_schema
 
 
 class TestMetadataModel:
+    def test_version_pattern(self):
+        # Valid versions
+        valid_versions = [
+            "1.0",
+            "0.1.0",
+            "1.0.0.dev1",
+            "2.3.4+local",
+            "3.4.5.dev6+g61d5f18",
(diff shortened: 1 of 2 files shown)


tests/test_agnostic/test_metadata_schema.py
```diff
@@ -1,36 +1,8 @@
...
-        versions = ["0.0", "0.1"]
-        regex = _get_versions_regex(versions)
-        assert re.match(regex, version_str) is None
-
-    def test_invalid_prefix_regex(self):
-        version_str = "9810.0.1"
-        versions = ["0.0", "0.1"]
-        regex = _get_versions_regex(versions)
-        assert re.match(regex, version_str) is None
+from geff.metadata_schema import Axis, GeffMetadata, write_metadata_schema
```
cmalinmayor, line 5: This is great! My only request would be to add back a few tests to show (how) the new regex works. I do not think in regex 😆 so I would want to verify it anyways

cmalinmayor (changes requested):
Just one request for a quick couple of tests of the regex (a couple that work, some that fail) to illustrate what is allowed for version and we are good to go! Thank you 🎉

cmalinmayor:
Ooh can you also remove the pyyaml dependency from the pyproject.toml? I'm 99% sure we don't need it anymore @yfukai

yfukai (commented):
> This is great! My only request would be to add back a few tests to show (how) the new regex works. I do not think in regex 😆 so I would want to verify it anyways 

Done!

cmalinmayor (approved):
🎉 wonderful thank you!

PR #161, opened 2025-07-19
WIP: flexible fixture to create dummy graph with full freedom 

TeunHuijben:
# Proposed Change
Flexible fixture to create dummy graph with full freedom over dtype/nodes/edges/numbers/props/etc

# Types of Changes
What types of changes does your code introduce?
- New feature or enhancement
- Tests

Which topics does your change affect?
- testing

# Checklist
Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read the [developer/contributing](https://github.com/live-image-tracking-tools/geff/blob/main/CONTRIBUTING) docs.
- [ ] I have added tests that prove that my feature works in various situations or tests the bugfix (if appropriate).
- [ ] I have checked that I maintained or improved code coverage.
- [ ] I have written docstrings and checked that they render correctly.

## If you changed the specification
- [ ] I have checked that any validation functions and tests reflect the changes.
- [ ] I have updated the GeffMetadata and the json schema using `pixi run update-schema` if necessary.
- [ ] I have updated docs/specification.md to reflect the change.
- [ ] I have updated implementations to reflect the change. (This can happen in separate PRs on a feature branch, but must be complete before merging into main.)

## If you have added or changed an implementation
- [ ] I wrote tests for the new implementation using standard fixtures supplied in conftest.py.
- [ ] I updated pyproject.toml with new dependencies if needed.
- [ ] I added a function to tests/bench.py to benchmark the new implementation.

# Further Comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #152, opened 2025-07-19
Common validation functions that aren't mandatory, but would be reused by many tools


msschwartz21:
Maybe live in a separate repo e.g. geff-validators
Such as
- Make sure that seg ids point a real object in the segmentation
- Track ids match whatever convention we agree on in #151

georgeoshardo:
@fjorka I have just added a util to calculate tracklet IDs in https://github.com/live-image-tracking-tools/geff/tree/calculate-tracklets-util. Your validator validates the tracklets. 
@msschwartz21 would this kind of thing be useful? I think people will be calculating this property a lot.

bentaculum:
@msschwartz21 I would also be interested to put functions that do structure validation (level 2) in the right place, is there already a draft for this?

msschwartz21:
> [@fjorka](https://github.com/fjorka) I have just added a util to calculate tracklet IDs in https://github.com/live-image-tracking-tools/geff/tree/calculate-tracklets-util. Your validator validates the tracklets. [@msschwartz21](https://github.com/msschwartz21) would this kind of thing be useful? I think people will be calculating this property a lot.

Yes we should have this in geff probably in geff.utils

msschwartz21:
Update: I did open a draft PR #204  to start working on this, but I haven't rearranged modules at all or started writing the logic for calling validation functions based on the contents of the metadata. So it's very very early. If anyone has gotten farther than I have please feel free to continue surging ahead.

cmalinmayor:
#180 partially addresses this, #213 will finish it

PR #180, opened 2025-07-21
Validator functions for geff data

fjorka:
# Proposed Change
This branch contains validators for checking different aspects of data in geff in response to #152.
It also provides a generator of correct and incorrect geff graphs to be replaced with a general approach in #161.

# Types of Changes
- New feature or enhancement
- Tests

- Specification
- `networkx` implementation

# Checklist

- no change

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,41 @@
+changelog:
+  categories:
+    - title: '✨ Features'
+      labels:
+        - ':star: feature'
+        - 'enhancement'
+    - title: ':clipboard: Specification'
+      labels:
+        - ':clipboard: specification'
+        - ':label: metadata'
+    - title: '🚀 Performance'
+      labels:
+        - ':rocket: performance'
+    - title: ':car: implementation'
+      labels:
+        - ':car: implementation'
+    - title: ':mag_right: validation'
+      labels:
+        - ':mag_right: validation'
+    - title: ':steam_locomotive: interoperability'
+      labels:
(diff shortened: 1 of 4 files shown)


src/geff/validators/validators.py
```diff
@@ -0,0 +1,149 @@
...
+        tuple[bool, list[str]]:
+            - is_valid (bool): True if all tracklets are valid, otherwise False.
+            - errors (list[str]): List of error messages for invalid tracklets.
+    """
+    errors = []
+
+    # if hasattr(edge_ids, "tolist"):
+    #     edges = edge_ids.tolist()
+    # else:
+    #     edges = list(edge_ids)
```
msschwartz21, line 106: 👀 ?

fjorka: Gone. Was fighting with types at some point...

cmalinmayor (commented):
I only skimmed the final iteration but this makes me so happy 🤩 thank you @fjorka !!!

tlambert03:
quick note on using `np.typing.ArrayLike` as a type hint.  it implies anything that *can be cast to* a numpy array... not a numpy array itself.  So you can't directly use numpy array attrs/methods like `.dtype` or `.shape` on an `ArrayLike` without first calling `np.asarray()` on it.  It *is* a useful way to say "you can give me anything that generally works with numpy" (dask, jax, xarray, zarr, etc...) ... but you can't assume anything about it other than the ability to cast it

PR #185, opened 2025-07-21
Add dataframe interops

JoOkuma:
# Proposed Change
Add API to convert nodes and edges to pandas dataframe

# Types of Changes
- New feature or enhancement

# Checklist

- [X] I have read the [developer/contributing](https://github.com/live-image-tracking-tools/geff/blob/main/CONTRIBUTING) docs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ docs = [
     "typer>=0.14.0,<0.17.0",
     "ruff>=0.12.5",
     "typing-extensions>=4.14.1",
-    "mkdocs-autorefs>=1"
+    "mkdocs-autorefs>=1",
     "griffe-fieldz",
 ]
 bench = [


src/geff/interops/_dataframe.py
```diff
@@ -0,0 +1,75 @@
...
+            for i in range(remaining_dims[0]):
+                new_columns.append(f"{c}_{i}")
+        columns = new_columns
+
+    df.columns = columns
+
+    return df
+
+
+def geff_to_dataframes(
```
cmalinmayor, line 51: `geff_from_dataframes` as well?

msschwartz21: I think that we shouldn't implement the from dataframes part. Initially I like the symmetry, but then we have to make all sorts of assertions about what is in the csv which is sort of what we're trying to avoid with geff in the first place.

msschwartz21:
@cmalinmayor Separate design question for this PR: should the function return dataframes or write them to disk as csv? All the other functions in the convert module go to disk so I'm somewhat tempted to go that way instead of returning the in memory option.

pyproject.toml
```diff
@@ -44,12 +44,13 @@ ctc = [
     "tifffile>=2024.10",
 ]
 rx = ["rustworkx>=0.16.0"]
+df = ["pandas>=2","pandas>=2.3; python_version >= '3.12'"]
```
msschwartz21, line 47: @tlambert03 I got into a bit of a fight with CI trying to get the lowest-direct versions of the tests passing with pandas. Do these pins seem reasonable to you?

tlambert03: ah... you were probably realizing that you need to be a bit more involved (than you'd like to have to be) in picking the lowest supported pandas based on which wheels are on pypi right?  There's a uv issue for that somewhere... will try to find.  But I often just end up doing what you've done here

msschwartz21: Yup that is exactly what happened. I'm fine with leaving it as is knowing that you have gone down a similar rabbit hole before 😂 

Good call about df vs. pandas. I don't have any intention of supporting a different pandas-like backend so I'll switch it to pandas

PR #197, opened 2025-07-22
TrackMate XML to GEFF

lxenard:
# Proposed Change
Add a converter to convert TrackMate XML to GEFF.
For now, it only extracts the graph data, including nodes and edges properties, and relies on write_nx to build a valid GEFF.

# Types of Changes
- New feature or enhancement
- Tests (incoming)

- io

# Checklist

- [x] I updated pyproject.toml with new dependencies if needed.

# Further Comments
The actual implementation relies on iterparse from lxml (not part of standard library). It means that the converter iteratively parses the XML, line by line. At no point is the XML completely loaded into memory. So, nice for crazy huge XMLs but probably comes at the cost of efficiency.

What is missing before being ready for review:
- proper tests but it's a pain to simulate XMLs (I just tested on a random XML: the graph is correct and the GEFF valid (thank you Talley!))
- docstrings consistency

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ classifiers = [
 
 [dependency-groups]
 test = ["pytest>=8.3.4", "pytest-cov>=6.2"]
+trackmate = ["lxml"]
 dev = [
     { include-group = "test" },
     "mypy",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 
 [dependency-groups]
 test = ["pytest>=8.3.4", "pytest-cov>=6.2"]
-trackmate = ["lxml"]
+trackmate = ["lxml", "typer"]
 dev = [
     { include-group = "test" },
     "mypy",
--- a/src/geff/interops/trackmate.py
+++ b/src/geff/interops/trackmate.py
@@ -1,0 +1,58 @@
+from pathlib import Path
+
+import lxml
+import typer
+
+
+def from_trackmate_xml_to_geff(
+    xml_path: Path,
+    geff_path: Path,
+    tczyx: bool = False,
+    overwrite: bool = False,
+) -> None:
+    """
+    Convert a TrackMate XML file to a GEFF file.
+
+    Args:
+        xml_path: The path to the TrackMate XML file.
+        geff_path: The path to the GEFF file.
+        tczyx: Expand data to make it (T, C, Z, Y, X) otherwise it's (T,) + Frame shape.
+        overwrite: Whether to overwrite the GEFF file if it already exists.
+    """
+    pass
+
+
+def from_trackmate_xml_to_geff_cli(
+    xml_path: Path,
+    geff_path: Path,
+    tczyx: bool = False,
+    overwrite: bool = False,
+) -> None:
+    """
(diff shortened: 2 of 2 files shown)


src/geff/interops/trackmate_xml.py
```diff
@@ -0,0 +1,595 @@
...
+            assert exit_node["TRACK_ID"] == current_track_id, err_msg
+    finally:
+        element.clear()
+
+
+def _build_tracks(
+    iterator: ET.iterparse,
+    ancestor: ET._Element,
+    attrs_md: dict[str, dict[str, str]],
+    graph: nx.Graph,
```
cmalinmayor, line 325: I think now that @melisande-c has improved our reference implementation, we don't need to go through networkx anymore. We can just create arrays of ids and values and directly write those. Perhaps you and Milly can get together today and clean this up? I am super excited about merging the trackmate XML converter so I am also happy to merge anything working and clean it up for you later!

lxenard: Yes I was thinking that yesterday, it would be cleaner and faster!

src/geff/interops/trackmate_xml.py
```diff
@@ -0,0 +1,583 @@
...
+def from_trackmate_xml_to_geff(
+    xml_path: Path | str,
+    geff_path: Path | str,
+    discard_filtered_spots: bool = False,
+    discard_filtered_tracks: bool = False,
+    overwrite: bool = False,
+    zarr_format: Literal[2, 3] = 2,
+) -> None:
+    """
+    Convert a TrackMate XML file to a GEFF file.
```
msschwartz21, line 509: Your comments and documentation through this PR are amazing!

I'm just wondering if there is more trackmate specific information that might be needed here with regards to how different special properties are handled. I can't really judge myself because I'm not familiar enough with trackmate xml

tests/test_interops/test_trackmate_xml.py
```diff
@@ -0,0 +1,583 @@
...
+
+import networkx as nx
+import networkx.algorithms.isomorphism as iso
+import pytest
+from lxml import etree as ET
+
+import geff.interops.trackmate_xml as tm_xml
+
+
+def is_equal(obt, exp):
```
msschwartz21, line 12: This will be useful for other people and can live in a more central place. I would put it either in utils or testing.utils if that exists.

tests/test_interops/test_trackmate_xml.py
```diff
@@ -0,0 +1,583 @@
...
+        "Not adding this node to the graph."
+    )
+    with pytest.warns(UserWarning, match=msg):
+        tm_xml._add_all_nodes(it, element, {}, nx.Graph())
+
+
+# Raises:
+#     AssertionError: If the 'TRACK_ID' attribute of either the source or target node
+#         does not match the current track ID, indicating an inconsistency in track
+#         assignment.
```
msschwartz21, line 328: 👀 ?

msschwartz21 (changes requested):
I think my big testing request is to add something that actually writes all the way to disk and calls validate on the resulting file. Otherwise this PR is amazing and thank you for all of the careful commenting and documentation!

msschwartz21 (approved):
🎉

PR #199, opened 2025-07-22
Automatic string conversion to bytes

JoOkuma:
# Proposed Change
Casts strings to bytes so it's compatible with java's zarr. 

# Types of Changes
- New feature or enhancement

- base writer

# Checklist

- [X] I have read the [developer/contributing](https://github.com/live-image-tracking-tools/geff/blob/main/CONTRIBUTING) docs.
- [X] I have added tests that prove that my feature works in various situations or tests the bugfix (if appropriate).
- [X] I have checked that I maintained or improved code coverage.
- [X] I have written docstrings and checked that they render correctly by looking at the docs preview (link left as a comment on the PR).

# Further Comments

Should I cast it back to `string` when reading? I think this could be problematic because the bytes might be something else.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -145,6 +145,7 @@ filterwarnings = [
     "ignore:'cgi' is deprecated and slated for removal in Python 3.13",
     # if needed, add global ignores here, for example:
     # "ignore:Some message:SomeCategory",
+    "ignore:Property '.*' is a string array. Automatically casting it to bytes:UserWarning",
 ]
 
 
--- a/src/geff/dict_representation.py
+++ b/src/geff/dict_representation.py
@@ -29,6 +29,7 @@ class PropDictNpArray(TypedDict):
 class PropDictZArray(TypedDict):
     values: zarr.Array
     missing: NotRequired[zarr.Array]
+    dtype: NotRequired[str]
 
 
 # Intermediate dict format that can be injested to different backend types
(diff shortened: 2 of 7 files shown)


cmalinmayor:
Yes, read it back in as strings. Minimally, store the string properties in the geff metadata so you know which ones to cast back. Even better, put it in the .zattrs of the property :)

JoOkuma:
Hi @cmalinmayor, I'm casting strings back to their original format

cmalinmayor (approved):
Can you either add a sentence in the spec about this, or make an issue to add a sentence in the spec about this? The code looks good, I am happy to merge either way.

cmalinmayor (changes requested):
Although, now that I am looking at other PRS, how does this interact with #206 and #207? #207 specifically does a different string to byte conversion, so I am revoking my approval until we pick one

JoOkuma:
An update of our discussion, @ksugar is looking into switching to n5-zarr because it supports `S` dtype, and there is work to support `U` dtype, see  https://github.com/saalfeldlab/n5-zarr/issues/74

#206 and this PR will wait to see if n5-zarr works

cc: @cmalinmayor and @msschwartz21

cmalinmayor:
Closing because Java will support string with switch to n5 reader :)

cmalinmayor:
JK we will do this with an explicit encoding, and store the dtype (and encoding) in the metadata of the property (Re: Jan and Ko discussing

cmalinmayor:
Actually closing in favor of #251

Issue #178, opened 2025-07-21
TODO: cli validator


tlambert03:
would be extremely convenient if there was a cli validation tool:

```sh
uvx geff validate path/to/project.zarr
```

that's something I'd be willing to help with

PR #200, opened 2025-07-22
Command line tool

yfukai:
# Proposed Change
Adding `uvx` callable command line tool, solving #178.

# Types of Changes
- New feature or enhancement

# Checklist


yfukai:
Maybe I should add docs..

msschwartz21:
Caveat: I haven't used uvx before; however, I tried the most naive thing to test the cli and it didn't work...

```
schwartzm10@schwartzm10-lm1 geff % uvx geff --help
Installed 16 packages in 33ms
The executable `geff` was not found.
warning: Package `geff` does not provide any executables.
```

msschwartz21:
I was going to play with it more before raising this issue but I think zarr 2/3 may be a bit tricky. If uvx installs zarr 2 then we won't be able to validate zarr 3 geffs. We could solve this by just defaulting to zarr 3 for the purposes of this cli tool.

yfukai:
Oh yeah you need to publish the package to run this. I used `uv` since it was mentioned in the original post by @tlambert03 but don't have a strong opinion on the choice of the tool.
For local test, you can do something like
```bash
python -m build
uvx --with zarr==3.1.0 --from dist/geff-0.3.1.dev28+g2e01560.d20250722-py3-none-any.whl validate /Users/fukai/projects/geff-benchmark/data/cell_segmentation.zarr/tracks.geff
```

yfukai:
- I pinned the zarr version in the script header (only for the cli script).
- I added a short doc for using command-line tool in the dev environment. (I couldn't find an easier way unfortunately)

cmalinmayor:
> In `traccuracy` I was able to do some minimal testing of a cli tool using typer. https://github.com/live-image-tracking-tools/traccuracy/blob/main/tests/test_cli.py
> 
> This would require replacing argparse with typer, but personally I like typer better and we already have it as a dependency.

I believe @JoOkuma has been using typer for `geff` cli things, so perhaps he can consult on this today.

yfukai:
Sure, I'll try `typer`!

yfukai:
I believe it's done, and would be happy to have your reviews @tlambert03 and @msschwartz21!

msschwartz21 (approved):
Looks good! I tested locally following your docs (thank you :joy:) and everything worked perfectly!

PR #222, opened 2025-07-23
Only check `path.exists` for local paths

DragaDoncila:
# Proposed Change
The current validator precludes the handling of remote GEFFs because it explicitly checks `path.exists`, which will always return `False` for a remote GEFF.

This PR omits that check for URLs that are parsed as remote using the builtin `urllib` library.

# Types of Changes
-  enhancement
- validators

# Checklist


cmalinmayor (approved):
🚜 🚋 📐

PR #226, opened 2025-07-23
Chore: install mypy pydantic plugin and pydantic additional dep

melisande-c:
# Proposed Change
Pydantic types weren't being properly checked by mypy. To fix this I add the pydantic mypy plugin to `pyproject.toml` and add pydantic as an additional dependency to `.pre-commit-config.yaml`. There was one error because it seems even the pydantic plugin cannot tell if a value is added in a before model validation. I have fixed this by setting the geff version through the `Field` default instead.

I tried to also add numpy as an additional dependency but there were 71 errors... so I commented it out.

# Types of Changes

- type checking

# Checklist


melisande-c:
Hmm ok, I didn't see this past discussion about the geff_version metadata #107

msschwartz21:
I'm turning this into a draft until we figure out how to keep the geff version field as required in the schema. I'm not opposed to potential just doing `# type: ignore` if that's the only way to get it to work.

src/geff/metadata_schema.py
```diff
@@ -190,7 +190,7 @@ class GeffMetadata(BaseModel):
     )
 
     geff_version: str = Field(
-        ...,
+        default=version("geff"),
```
tlambert03, line 193: this is a good pattern.  all calls to `Field()` with a positional default first argument should be replaced with `default=` (such as `sphere`, `track_node_props`, etc...), otherwise some type checkers interpret it as a required argument

PR #243, opened 2025-07-24
Reorganize cli to contain conversion functions in cli file

### LINKED ISSUE #220: Consolidate cli tools
Currently the ctc cli is buried inside the interops/ctc module

msschwartz21:
# Proposed Change
- Closes #220 
- Improves handling of imports for conversion functions which require optional dependencies

Currently there still isn't any testing of raising errors when optional imports are missing, but that's outside the scope of this PR. 

# Types of Changes

# Checklist


src/geff/_cli.py
```diff
@@ -27,5 +32,106 @@ def info(input_path: str = typer.Argument(..., help="Path to the GEFF file")):
     print(metadata.model_dump_json(indent=2))
 
 
+@app.command()
+def convert_ctc(
```
TeunHuijben, line 36: out of curiosity, how come that the function is called `convert_ctc`, but on the command-line we have to use `convert-ctc` (middle dash vs lower dash)? Does the documentation contain info on how to use the cli, apart from `geff --help` in the terminal?

src/geff/_cli.py
```diff
@@ -27,5 +32,106 @@ def info(input_path: str = typer.Argument(..., help="Path to the GEFF file")):
...
+    ctc_path: Annotated[
+        Path,
+        typer.Argument(help="The path to the directory containing ctc tracks", show_default=False),
+    ],
+    geff_path: Annotated[
+        Path, typer.Argument(help="Path to save the output geff", show_default=False)
```
TeunHuijben, line 42: I think the `geff_path` variable argument can be explained a bit more clearly. For example: 

"Path to save the output geff, including the geff directory name (eg. `~/folder/folder/tracks.geff`)

I got confused when trying it out, because when I used `~/Desktop/` as `geff_path`, I was expecting it to save the geff to my desktiop, but in stead,  the geff was created as `~/Desktop.geff`.

src/geff/_cli.py
```diff
@@ -27,5 +32,106 @@ def info(input_path: str = typer.Argument(..., help="Path to the GEFF file")):
...
+            overwrite=overwrite,
+        )
+
+
+@app.command()
+def convert_trackmate_xml(
+    xml_path: Annotated[
+        Path, typer.Argument(help="The path to the TrackMate XML file", show_default=False)
+    ],
+    geff_path: Annotated[
```
TeunHuijben, line 104: same here

src/geff/_cli.py
```diff
@@ -27,5 +32,106 @@ def info(input_path: str = typer.Argument(..., help="Path to the GEFF file")):
     print(metadata.model_dump_json(indent=2))
 
 
+@app.command()
+def convert_ctc(
+    ctc_path: Annotated[
+        Path,
+        typer.Argument(help="The path to the directory containing ctc tracks", show_default=False),
```
TeunHuijben, line 39: Can we be bit more explicit here, by either mentioning the `TRA` folder, or adding: 

(eg. `data/01_GT/TRA/`) ?

TeunHuijben (approved):
Looks good! I tested the cli functions on an `.xml` and an example CTC file, and both work perfectly 🙌 . I only have minor comments regarding the cli argument descriptions.

src/geff/_cli.py
msschwartz21: We do have a page in the docs that documents the cli: http://liveimagetrackingtools.org/geff/pr-preview/pr-243/command-line-tools/. Let me know if you think that is sufficient.

The pattern of converting _ to - in a cli context is pretty standard. Mostly python likes _ and the convention for command line is - so the conversion happens automatically with the tool we are using (typer).

msschwartz21:
@tlambert03 I think I broke something when I merged in main, but I can't figure out why it's passing in some tests, but not others. Can you take a look?

src/geff/interops/__init__.py
```diff
@@ -0,0 +1,33 @@
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from .ctc import ctc_tiffs_to_zarr, from_ctc_to_geff
+    from .trackmate_xml import from_trackmate_xml_to_geff
```
tlambert03, line 5:
```suggestion
from geff.interops.trackmate_xml import from_trackmate_xml_to_geff

if TYPE_CHECKING:
    from .ctc import ctc_tiffs_to_zarr, from_ctc_to_geff
```

no longer anything in trackmate_xml that needs an extra

src/geff/interops/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+            raise ImportError("install with geff[ctc] to use from_ctc_to_geff") from e
+    if name == "from_trackmate_xml_to_geff":
+        try:
+            from geff.interops.trackmate_xml import from_trackmate_xml_to_geff
+
+            return from_trackmate_xml_to_geff
+        except ImportError as e:
+            raise ImportError(
+                "install with geff[trackmate_xml] to use from_trackmate_xml_to_geff"
+            ) from e
```
tlambert03, line 33:
```suggestion
def __getattr__(name: str) -> Any:
    if name == "ctc_tiffs_to_zarr":
        from geff.interops.ctc import ctc_tiffs_to_zarr

        return ctc_tiffs_to_zarr
    if name == "from_ctc_to_geff":
        from geff.interops.ctc import from_ctc_to_geff

        return from_ctc_to_geff
```

these catches are already in `ctc`

src/geff/_cli.py
```diff
@@ -1,31 +1,159 @@
...
+        ..., help="Path to the GEFF folder, e.g. data.zarr/tracks.geff"
+    ),
+):
     """Display information about a GEFF file."""
     metadata = GeffMetadata.read(zarr.open(input_path, mode="r"))
     print(metadata.model_dump_json(indent=2))
 
 
+@app.command()
+def convert_ctc(
```
tlambert03, line 44: it would be nifty (if perhaps require a bit more "magic") to have one `geff convert` ... rather than a different `convert_x` for each thing you intend to support?  seems to me like it should be possible to have a `guess_format(path)` helper function that takes the input and infers the format?

on the other hand... if you know that each conversion format is going to require a lot of format-specific arguments, maybe not...

src/geff/_cli.py
```diff
@@ -1,31 +1,159 @@
...
 from . import utils
 from .metadata_schema import GeffMetadata
 
+if TYPE_CHECKING:
+    from typing import Literal
+
 app = typer.Typer(help="GEFF Command Line Interface")
 
 logging.basicConfig(stream=sys.stderr, level=logging.WARNING, format="%(levelname)s: %(message)s")
 logging.captureWarnings(True)
```
tlambert03, line 18: i know it's not from this PR... but do you know why we're configuring/using logging here?

src/geff/_cli.py
```diff
@@ -1,31 +1,159 @@
...
+    overwrite: Annotated[
+        bool, typer.Option(help="Whether to overwrite the GEFF file if it already exists.")
+    ] = False,
+    zarr_format: Annotated[int, typer.Option(help="The version of zarr to write.")] = 2,  # type: ignore
+    # because of Typer not supporting Literal types
+) -> None:
+    """
+    Convert a TrackMate XML file to a GEFF file.
+    """
+    from geff.interops import from_trackmate_xml_to_geff  # noqa: I001 import at call time to avoid optional dependency issues
```
tlambert03, line 146:
```suggestion
    from geff.interops import from_trackmate_xml_to_geff 
```

(unneeded noqa ... ruff doesn't mind imports inside functions, they're fine... and absolutely standard in this scenario)

tests/test_interops/test_ctc.py
```diff
@@ -64,44 +63,24 @@ def create_mock_data(
 
 
 @pytest.mark.parametrize(
-    "cli,is_gt,tczyx",
-    list(itertools.product([False, True], [True, False], [True, False])),
+    "is_gt,tczyx",
```
tlambert03, line 66: i know it's not from this PR... but it would be a bit more idiomatic/clear to use two params, instead of itertools:

```python
@pytest.mark.parametrize("is_gt", [True, False])
@pytest.mark.parametrize("tczyx", [True, False])
```

tests/test_cli.py
```diff
@@ -37,3 +40,47 @@ def test_info_command_prints_metadata(example_geff_path):
 def test_main_invalid_command():
     result = runner.invoke(app, ["invalid"])
     assert result.exit_code != 0
+
+
+@pytest.mark.parametrize(
+    "is_gt,tczyx",
+    list(itertools.product([True, False], [True, False])),
+)
```
tlambert03, line 48:
```suggestion
@pytest.mark.parametrize("is_gt", [True, False])
@pytest.mark.parametrize("tczyx", [True, False])
```

same as above (pytest naturally gives you a cartesian product if you use two params)

src/geff/_cli.py
msschwartz21: I'm game to try to have a general convert that just uses all defaults if that makes sense. Need to think about it a bit more closely

src/geff/_cli.py
msschwartz21: No clue. Happy to pull it out if it's not useful

src/geff/interops/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+            from geff.interops.ctc import from_ctc_to_geff
+
+            return from_ctc_to_geff
+        except ImportError as e:
+            raise ImportError("install with geff[ctc] to use from_ctc_to_geff") from e
+    if name == "from_trackmate_xml_to_geff":
+        try:
+            from geff.interops.trackmate_xml import from_trackmate_xml_to_geff
+
+            return from_trackmate_xml_to_geff
```
cmalinmayor, line 29: I think trackmate XML is no longer having optional dependencies?

cmalinmayor:
I was about to try and fix this but @TeunHuijben beat me to it 😄 thank you!!

PR #247, opened 2025-07-25
TrackMate XML metadata to geff

lxenard:
# Proposed Change
Parses all the TrackMate XML metadata and saves it into the geff. Up to date with the current specification!

# Types of Changes
- New feature or enhancement
- Tests

- interops

# Checklist

### DIFF
--- a/src/geff/interops/trackmate_xml.py
+++ b/src/geff/interops/trackmate_xml.py
@@ -100,13 +100,13 @@ def _get_units(
     if "spatialunits" not in units:
         warnings.warn(
             "No space unit found in the XML file. Setting to 'pixel'.",
-            stacklevel=2,
+            stacklevel=4,
         )
         units["spatialunits"] = "pixel"  # TrackMate default value.
     if "timeunits" not in units:
         warnings.warn(
             "No time unit found in the XML file. Setting to 'frame'.",
-            stacklevel=2,
+            stacklevel=4,
         )
         units["timeunits"] = "frame"  # TrackMate default value.
     element.clear()  # We won't need it anymore so we free up some memory.
@@ -154,6 +154,7 @@ def _convert_attributes(
     attrs: dict[str, Any],
     attrs_metadata: dict[str, dict[str, str]],
     attr_type: str,
+    stack_level: int,
 ) -> None:
     """
     Convert the values of the attributes from string to the correct data type.
@@ -166,6 +167,8 @@ def _convert_attributes(
         attrs_metadata (dict[str, dict[str, str]]): The attributes metadata containing
         information on the expected data types for each attribute.
         attr_type (str): The type of the attribute to convert (node, edge, or lineage).
+        stack_level (int): The stack level for warnings, used to indicate where
+            the warning originated.
 
     Raises:
         ValueError: If an attribute value cannot be converted to the expected type.
@@ -194,7 +197,7 @@ def _convert_attributes(
         else:
             warnings.warn(
                 f"{attr_type.capitalize()} attribute {key} not found in the attributes metadata.",
-                stacklevel=2,
+                stacklevel=stack_level,
             )
 
 
@@ -261,7 +264,7 @@ def _add_all_nodes(
             # on these attributes type as defined in the attributes
             # metadata (attribute `isint`).
             attrs = deepcopy(element.attrib)
-            _convert_attributes(attrs, attrs_md, "node")
+            _convert_attributes(attrs, attrs_md, "node", 5)
 
             # The ROI coordinates are not stored in a tag attribute but in
             # the tag text. So we need to extract then format them.
@@ -282,7 +285,7 @@ def _add_all_nodes(
                 warnings.warn(
                     f"No key 'ID' in the attributes of current element "
                     f"'{element.tag}'. Not adding this node to the graph.",
-                    stacklevel=2,
+                    stacklevel=4,
                 )
             finally:
                 element.clear()
@@ -320,7 +323,7 @@ def _add_edge(
         UserWarning: If an edge cannot be added due to missing required attributes.
     """
     attrs = deepcopy(element.attrib)
-    _convert_attributes(attrs, attrs_md, "edge")
+    _convert_attributes(attrs, attrs_md, "edge", 6)
     try:
         entry_node_id = int(attrs["SPOT_SOURCE_ID"])
         exit_node_id = int(attrs["SPOT_TARGET_ID"])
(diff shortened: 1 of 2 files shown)


lxenard:
Noob question: do I need to tag you @msschwartz21 @cmalinmayor when a PR switches to 'ready for review', or do you receive a notification automatically?

msschwartz21:
> Noob question: do I need to tag you @msschwartz21 @cmalinmayor when a PR switches to 'ready for review', or do you receive a notification automatically?

If we do get notification when it's switched to ready to review, it's not particularly noticeable 😅 . I'd appreciate it if you request us as reviewers because I keep an eye out for PRs that are waiting on my review.

src/geff/interops/trackmate_xml.py
```diff
@@ -462,55 +482,450 @@ def _parse_model_tag(
...
-        if element.tag == "FilteredTracks" and event == "start":
-            # Removal of filtered tracks.
-            id_to_keep = _get_filtered_tracks_ID(it, element)
-            if discard_filtered_tracks:
-                to_remove = [n for n, t in graph.nodes(data="TRACK_ID") if t not in id_to_keep]
-                graph.remove_nodes_from(to_remove)
-
-        if element.tag == "Model" and event == "end":
-            root.clear()
+    with open(xml_path, "rb") as f:
```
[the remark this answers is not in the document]
msschwartz21: Is it just showing as a big diff because everything got indented?

src/geff/interops/trackmate_xml.py
```diff
@@ -462,55 +482,450 @@ def _parse_model_tag(
...
+                trackmate_metadata["DisplaySettings"], encoding="utf-8"
+            ).decode(),
+        },
+    }
+
+    return GeffMetadata(
+        axes=[
+            Axis(name="POSITION_X", type="space", unit=units.get("spatialunits", "pixel")),
+            Axis(name="POSITION_Y", type="space", unit=units.get("spatialunits", "pixel")),
+            Axis(name="POSITION_Z", type="space", unit=units.get("spatialunits", "pixel")),
```
msschwartz21, line 851: Mostly a curiosity question, but does trackmate always assume a z dimension? How would 2d data be handled?

src/geff/interops/trackmate_xml.py
```diff
@@ -462,55 +482,450 @@ def _parse_model_tag(
...
+    removed_props = []
+    for prop, data_iter in zip(props_to_check, iterators, strict=True):
+        if not any(prop in data for data in data_iter):  # early termination when prop found
+            metadata_dict.pop(prop)
+            removed_props.append(prop)
+
+    if removed_props:
+        plural1 = "ies were" if len(removed_props) > 1 else "y was"
+        plural2 = "they are" if len(removed_props) > 1 else "it is"
+        logging.info(
```
msschwartz21, line 896: This might be better as a warning? We're not using logging heavily in this library and I might be concerned that the message would be lost to the average user.

src/geff/interops/trackmate_xml.py
```diff
@@ -462,55 +482,450 @@ def _parse_model_tag(
...
+            f"The following {component_type} property{plural1} removed from the metadata "
+            f"because {plural2} not present in the data: {', '.join(removed_props)}."
+        )
+
+
+def _ensure_data_metadata_consistency(
```
msschwartz21, line 902: Just out of curiosity, why do you think this check is necessary?

src/geff/interops/trackmate_xml.py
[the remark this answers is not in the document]
lxenard: Yes it's just a change of indentation due to the addition of the context manager. I had to add it because the file was not closing properly otherwise.

src/geff/interops/trackmate_xml.py
```diff
@@ -462,55 +482,450 @@ def _parse_model_tag(
...
+        img_path (str | None): The path to the related image file.
+        trackmate_metadata (dict[str, dict[str, ET.Element]]): The TrackMate metadata extracted
+            from the XML file.
+        props_metadata (dict[str, Any]): The properties metadata extracted from the XML file.
+
+    Returns:
+        GeffMetadata: The constructed GEFF metadata object.
+    """
+    extra = {
+        "trackmate_version": _get_trackmate_version(xml_path),
```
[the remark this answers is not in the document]
lxenard: Ah yes, I forgot!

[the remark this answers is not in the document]
lxenard: In that case, we don't really need a provenance field since the presence of 'other_trackmate_metadata' already indicates that the data is coming from TM

src/geff/interops/trackmate_xml.py
lxenard: At first I put a warning, but had to replace it otherwise **test_from_trackmate_xml_to_geff** fails... I don't know why I didn't just account for the warnings in the test 🤷 Switching back to the warning!

src/geff/interops/trackmate_xml.py
lxenard: Actually it happens quite often that TM declares properties but do not compute them, so it's really needed to get a valid GEFF.

src/geff/interops/trackmate_xml.py
lxenard: So I had to add an ignore in the pyproject.toml to avoid failure for the CLI test.

lxenard:
The FakeTracks.xml should cover all types of data / metadata, I modified it manually with that intent. But tomorrow I will run the converter on a bunch of different files to be sure, and let you know! Thanks for the review!

src/geff/interops/trackmate_xml.py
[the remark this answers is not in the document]
msschwartz21: Yup I agree

src/geff/interops/trackmate_xml.py
msschwartz21: Hahah ok this is a classic cascade 

First things, uncaught warnings will cause tests to fail which is desirable so we don't to disable it. If there is an expected warning you add a context `with pytest.warns(UserWarning): ...` ideally with a match for the specific warning.

There is a way to get typo to stop fixing things. Let me look and I'll point you to the place to add the new ignore

msschwartz21: Add whatever you need typo to ignore here: https://github.com/live-image-tracking-tools/geff/blob/2fd8cfe4442c11c34a6c744cc7a3644c1570fb9c/pyproject.toml#L160

cmalinmayor:
^ Iconic commit message 😆 😭

lxenard:
At least now I know that I'm not completely crazy! 😆

lxenard:
So I've tried on different XMLs and everything runs fine. Of course if people starts to mess up manually with the XML, the converter is likely to crash, but in a normal use case it should run properly on every XML since TrackMate export to XML is very stable. If someone raises an issue about the converter, I would be happy to fix it! Just tag me or whatever so I'm aware of the issue.

PR #251, opened 2025-07-26
Implement vlen_props to support variable-length content

ksugar:
# Proposed Change
This PR implements support for reading and writing variable-length properties such as polygons. The variable length properties must be numpy arrays of the same number of dimensions and dtype, but can have any shape. There is a helper, `construct_var_len_props` that can convert a sequence of ArrayLike or None into this format.
Variable length properties are stored with an additional `data` entry that contains the data serialized into one long one-dimensional array of the shared dtype. The `values` array contains the offset and shape of the array for the corresponding node/edge. The `missing` array remains unchanged, and there are no restrictions on what is in the `values` array when the property is missing.

Strings are NOT encoded with this serialization scheme, but instead with the zarr-python implementation of the zarr-extensions specification for variable length strings.

# Types of Changes
- New feature or enhancement
- Documentation update
- Tests

- Specification

# Checklist

- [x] I wrote tests for the new implementation using standard fixtures supplied in conftest.py.

# Further Comments
Updated diagram coming soon

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -24,7 +24,7 @@ The nodes group will contain an `ids` array and optionally a `props` group.
 
 ### The `ids` array
 
-The `nodes\ids` array is a 1D array of node IDs of length `N` >= 0, where `N` is the number of nodes in the graph. Node ids must be unique. Node IDs must have an unsigned integer dtype. For large graphs, `uint64` might be necessary to provide enough range for every node to have a unique ID. In the minimal case of an empty graph, the `ids` array will be present but empty.
+The `nodes\ids` array is a 1D array of node IDs of length `N` >= 0, where `N` is the number of nodes in the graph. Node ids must be unique. Node IDs must have an integer dtype. For large graphs, `int64` might be necessary to provide enough range for every node to have a unique ID. In the minimal case of an empty graph, the `ids` array will be present but empty.
 
 ### The `props` group and `node property` groups
 
--- a/src/geff/validate/structure.py
+++ b/src/geff/validate/structure.py
@@ -138,8 +138,8 @@ def _validate_nodes_group(nodes_group: zarr.Group, metadata: GeffMetadata) -> No
     node_ids = expect_array(nodes_group, _path.IDS, _path.NODES)
 
     # Node ids must be uint dtype
-    if not np.issubdtype(np.dtype(node_ids.dtype), np.unsignedinteger):
-        raise ValueError("Node ids must have an unsigned integer dtype")
+    if not np.issubdtype(np.dtype(node_ids.dtype), np.integer):
+        raise ValueError("Node ids must have an integer dtype")
 
     id_len = node_ids.shape[0]
     node_props = expect_group(nodes_group, _path.PROPS, _path.NODES)
--- a/tests/test_validate/test_structure.py
+++ b/tests/test_validate/test_structure.py
@@ -103,12 +103,7 @@ class Test_validate_nodes_group:
 
     def test_ids_not_int(self, node_group, meta):
         node_group[_path.IDS] = node_group[_path.IDS][:].astype("float")
-        with pytest.raises(ValueError, match="Node ids must have an unsigned integer dtype"):
-            _validate_nodes_group(node_group, meta)
-
-        # Must be uint, not just int
-        node_group[_path.IDS] = node_group[_path.IDS][:].astype("int")
-        with pytest.raises(ValueError, match="Node ids must have an unsigned integer dtype"):
+        with pytest.raises(ValueError, match="Node ids must have an integer dtype"):
             _validate_nodes_group(node_group, meta)
 
     # Other cases are caught in tests for _validate_props_group


src/geff/typing.py
msschwartz21, line 1: @melisande-c Do you mind taking a look at the new typing patterns and reviewing?

ksugar:
@bogovicj this is the PR that is implementing serialized_props, which will be renamed to veln_props soon.

ksugar:
I will change the name of the group from `serialized_props` to `vlen_props`.

docs/specification.md
```diff
@@ -54,6 +54,11 @@ The `nodes\props` group is optional and will contain one or more `node property`
 
     When writing a graph with missing properties to the geff format, you must fill in a dummy value in the `values` array for the nodes that are missing the property, in order to keep the indices aligned with the node ids. 
 
+### The `vlen_props` group and `node vlen property` groups
+The `nodes\vlen_props` group is optional and will contain one or more `node vlen property` groups, each with a `values` array, a `slices` array and an optional `missing` array. These properties are stored as a list of arrays, where each element in the array can have a variable length. This is useful for properties that are not fixed-length, such as polygons or other complex shapes.
```
cmalinmayor, line 58: Ideally, the vlen_props would be as similar to normal properties as possible. I would encourage them to be within the same group. I would even rename the `slices` array to be the `values` so that we can re-use our validation functions. The proposed `values` can be something like `data`.

ksugar:
I’ve revised the schematic illustration of the `vlen_props`.
Could you let me know if it looks what you expect?

vlen_props
vlen_props_example

### Sample code

```python
>>> import zarr
>>> from zarr.storage import MemoryStore
>>>
>>> vlen_props = zarr.group(store=MemoryStore())
>>>
>>> vlen_props["values"] = np.arange(24)
>>> vlen_props["offset"] = np.array((0, 4, 7, 13, 16, 23))
>>> vlen_props["shape"] = np.array(
...     (
...         (2, 2, -1),
...         (3, 1, -1),
...         (6, -1, -1),
...         (3, 8, 2),
...         (2, 2, 2),
...         (0, 1, 2),
...     )
... )
>>> vlen_props["missing"] = np.array((False, False, False, True, False, False))
>>>
>>>
>>> def deserialize(vlen_props: zarr.Group, index: int) -> np.ndarray | None:
...     if "missing" in vlen_props and vlen_props["missing"][index]:
...         return None
...     if 0 in vlen_props["shape"][index]:
...         return np.empty(
...             vlen_props["shape"][index].tolist(),
...             dtype=vlen_props["values"].dtype,
...         )
...     shape_i = vlen_props["shape"][index][vlen_props["shape"][index] > 0]
...     return vlen_props["values"][
...         vlen_props["offset"][index]:vlen_props["offset"][index] + shape_i.prod()
...     ].reshape(shape_i)
...
>>> deserialize(vlen_props, 0)
array([[0, 1],
       [2, 3]])
>>> deserialize(vlen_props, 1)
array([[4],
       [5],
       [6]])
>>> deserialize(vlen_props, 2)
array([ 7,  8,  9, 10, 11, 12])
>>> deserialize(vlen_props, 3)
>>> deserialize(vlen_props, 4)
array([[[16, 17],
        [18, 19]],

       [[20, 21],
        [22, 23]]])
>>> deserialize(vlen_props, 5)
array([], shape=(0, 1, 2), dtype=int64)
```

ksugar:
@cmalinmayor thank you for your feedback.
I’ll update the naming once we reach a consensus.
If we decide that each property must have the same number of dimensions (or be `None`) and restrict storing empty arrays, we won't need to support `-1` or `0` as data values.

msschwartz21:
> If we decide that each property must have the same number of dimensions (or be `None`) and restrict storing empty arrays, we won't need to support `-1` or `0` as data values.

I think this is the correct approach.

> If we can smush together the shape and offset into a `values` array and then rename `values` to `data` I think we can re-use validation functions. It makes the nameing a little less clear in favor of code re-use, so I'm open to debate.

Ok it took me a second but I think I agree with this. Right now in our validation we always check that the length of values matches the number of nodes/edges so this approach would cause our validation to fail. If shape and offset are smushed into a single array that is named values, our standard validation will pass. I would also be ok with offset -> values and then shape being it's own array. The actual 1d flattened array can be called data and be exempt from validation. 

For strings we exclude the shape part and only have offset?

ksugar:
We need to have the `shape` for strings if we support unicode (non-ascii) characters, as the length in bytes may change after encoding.

ksugar:
@JoOkuma I broadly agree with you. From my reading of the specification, values may be present or missing, but if flagged as missing, they are ignored regardless of whether a value is stored. Still, the typical case is likely as you described.
https://liveimagetrackingtools.org/geff/latest/specification/#the-props-group-and-node-property-groups

mkitti:
The Zarr v3 sharding codec currently provides a mechanism to describe variable length chunks.

https://zarr-specs.readthedocs.io/en/latest/v3/codecs/sharding-indexed/index.html

Google has even been using it recursively.


The index for these variable length chunks is as follows and code for decoding this already exists in many Zarr implementations.

```
| chunk (0, 0)    | chunk (0, 1)    | chunk (1, 0)    | chunk (1, 1)    |          |
| offset | nbytes | offset | nbytes | offset | nbytes | offset | nbytes | checksum |
| uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint64 | uint32   |
```

While this proposal is at a higher level, there are some similarities. One noticeable difference is that the sharding codec allows for chunks to not be consecutive within the "data". There could be gaps between the data components, the chunks.

This allows for implementation specific optimizations. The trivial one is packed chunks with no spaces between them in order to minimize storage.

Another is equally spacing the chunks so that chunks could expand to some maximum size without having to rearrange the chunks or rewrite the entire data portion. This allows for individual components to be independently updated without interfering with the others.

While one could infer the property lengths from the shapes here, it is not clear if property values need to be packed together with no space between them.

The final component is the inclusion of a crc32c checksum on the offsets and nbytes to ensure integrity of the index. Corruption of the offset indices could lead to the effective loss of the entire property array

cmalinmayor:
Thanks @mkitti, that's good to know. I don't think we need this level of optimization (we aren't even really chunking,  and we don't need to support changes to the data), nor do we want to confine ourselves to zarr v3 right now.

mkitti:
You do not necessarily need to use Zarr 3 to reuse any of these concepts. The set of Zarr 2 codecs is not closed anyways, so sharding could and probably will be implemented in Zarr v2.

Simply clarifying here that the length of the data in the serialized array should be calculated from the shapes rather than from the differences in the offsets would be sufficient to allow gaps between variable length properties.

That would substantially increase the compatibility of this scheme with the concept of chunking itself since the raw chunks are typically of fixed size. Allowing gaps would allow for alignment of data with chunk boundaries. This would allow implementations to ensure that variable length property arrays are not split over multiple chunks unnecessarily.

mkitti:
Note that the Java supported encodings above are specific to the Oracle Java implementation.

The charsets required by specification are much more limited:

https://docs.oracle.com/javase/8/docs/api/java/nio/charset/Charset.html

UTF-32 was not introduced as a standard character set until 2023:

https://bugs.openjdk.org/browse/JDK-8310251

ksugar:
@mkitti thanks. So, shall we limit the supported encodings to the following set?

- US-ASCII
- ISO-8859-1
- UTF-8
- UTF-16BE
- UTF-16LE
- UTF-16

mkitti:
I would just support UTF-8.

UTF-16 is unnecessary and introduces endianess complications.

The important parts of ASCII and latin-1 are basically UTF-8.

https://utf8everywhere.org/

bogovicj:
As written, it is not forbidden to have two different values use overlapping sections, or to entirely reuse parts of the data array, and that could be useful, actually.  Do you want to allow this?

In this example, shape is always `(1,)`

```
data = [0, 1]

values = [[  
  0, 0, 1, 1
  1, 1, 1, 1 
]]
```

This stores the properties for 4 nodes, but with only two unique values.

cmalinmayor:
> ## overlapping / re-using the data array
> As written, it is not forbidden to have two different values use overlapping sections, or to entirely reuse parts of the data array, and that could be useful, actually. Do you want to allow this?

Sure, why not! 

> ### use "empty shape" to mean "missing"?
> If you don't need to distinguish between empty and missing, then a shape of 0 could be used to indicate "missing"
> 
> If you stick with having a "missing" array, then I imagine there would be no constraints at all on the corresponding offset and shape, right? As in, why not put `0`s or `-1`s there? @ksugar seems to [agree](https://github.com/live-image-tracking-tools/geff/pull/251#issuecomment-3134008117)

We will stick with having a "missing" array to mirror the structure of the other, non variable length properties, where we can't rely on reserving values to mean "missing". For other properties, we don't constrain what values you put in the values array when there is a corresponding "missing" index. I agree it makes sense to put 0s or -1s for the shape and offset values, but the spec won't force you to.

lxenard:
Please tag me when this PR is merged, I will then update the TrackMate XML to GEFF converter with the segmentation data 🐙

cmalinmayor:
@ksugar I can't tag you as a reviewer because this is technically your PR 😆 but if you could review that would be great! I think we need to update the diagram just a bit (no more None returned when you read a missing value -  read_arrays just gives you the values and missing arrays, and depending on which graph backend you use it does different things with that information). Relatedly, there is a helper to support writing from your old format (sequence of ArrayLike or None), but no convenience reader to put it back in that format - is that something you would need on the python side?

msschwartz21 (commented):
Would you mind updating the main text of the PR with a recap of what we settled on? That text goes into the commit message so it's helpful if its more descriptive.

src/geff/core_io/_utils.py
```diff
@@ -172,3 +178,71 @@ def setup_zarr_group(store: StoreLike, zarr_format: Literal[2, 3] = 2) -> zarr.G
...
+    all elements have the same number of dimensions. Casts all arrays into a common dtype
+    (if there is one). Turns the sequence into a numpy array with dtype object.
+
+    Args:
+        arr_seq (Sequence[ArrayLike | None]): A sequence of properties, with one entry
+            per node or edge. Missing values are indicated by None entries.
+
+    Returns:
+        PropDictNpArray: A standardized version of the input properties where all entries
+        are numpy arrays contained in an object array.
```
msschwartz21, line 229:
```suggestion
            are numpy arrays contained in an object array.
```
Fix docstring rendering

cmalinmayor: I just subconsciously fixed this when I was reading through the file to add the other docstring 😆 My brain knows what sphinx wants

src/geff/core_io/_serialization.py
```diff
@@ -0,0 +1,122 @@
...
+        data (NDArray): The 1D array containing the serialized property data.
+            expected shape is (total_data_length,).
+        index (int): The index of the property to deserialize.
+
+    Returns:
+        NDArray: The deserialized variable-length value.
+    """
+    if index < 0 or index >= values.shape[0]:
+        raise IndexError(f"Index {index} out of bounds for property data.")
+    # TODO: we could use the missing values to make an empty array, but it messes
```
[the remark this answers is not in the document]
cmalinmayor: After discussion, it's just more checking of things that aren't technically wrong right now, since we just ignore the values for the missing arrays. So, we decided to leave it for now.

cmalinmayor:
> > @msschwartz21 related question: for props metadata dtype validation, we currently only warn if the dtype isn't a valid numpy dtype, but then we use the dtype when reading... Should I change it to an error?
> 
> Yes that makes sense.
> 
I did change it to an error, but now passing in float16 data causes an error instead of just a warning. Should we just upcast to float32?

ksugar:
Sorry for the delayed review, but I’ve confirmed that the handling of vlen props is now clear and the requirements are met. Thank you for completing the implementation.

PR #252, opened 2025-07-27
codspeed benchmarks

tlambert03:
this is a WIP PR to update the benchmarking framework

- it cleans up the benchmark code a bit (in preparation for testing additional backends like write_rx... though there appears to be a bug in write_rx at the moment preventing successful benchmarking)
- it uses codspeed https://github.com/CodSpeedHQ/action ... this has benefits:
    - the current strategy only tests regressions on non-forks (which is quite problematic if you can't catch regressions from external contributors)
    - codspeed does a lot of things to ensure reproducibility in a ci environment (which is non-trivial, and harder than simply using pytest-benchmark on github actions)
    - codspeed stores records permanently and provides a dashboard to review records 
    - it lets you fail PRs based on thresholds (more than just posting a comment on non-forks)
    - it can help you debug regressions with [flame graphs and profiling](https://codspeed.io/blog/pinpoint-performance-regressions-with-ci-integrated-differential-profiling).


msschwartz21:
I'll deal with merge conflicts and also add that spatial graph dependency to bench group

tlambert03:
It’s ok, I’ll update shortly

tlambert03:
thanks @msschwartz21!  I ended up reverting back to your write-then-read strategy (rather than having `create_sg_graph`) ... I think yours was the cleaner pattern, even if it might take just a moment longer :)

tests/test_bench.py
```diff
@@ -0,0 +1,115 @@
...
+# ###########################   TESTS   ##################################
+
+READ_PATH: Mapping[Callable, Callable[[Path], tuple[Any, Any]]] = {
+    geff.read_nx: lambda path: geff.read_nx(path, validate=False),
+    geff.read_rx: lambda path: geff.read_rx(path, validate=False),
+    geff.read_sg: lambda path: geff.read_sg(path, validate=False),
+}
+
+
+@pytest.mark.parametrize("nodes", [500])
```
tlambert03, line 90: ok, main things to note here:

rather than having a module level `NODES` constant at 2000, I chose to parametrize each test.  Even though there is (Currently) a single parameter, this will make it so that the number of nodes is listed in the test name, and codspeed will know that if you change or add parameters later, they should be interpreted as different benchmarks.  In other words: try not to have hidden parameters that will affect the speed of the test.

Another thing to note is that I brought it down from 2000 to 500.  This is because benchmakrs actually *do* take longer when you run them on codspeed.  Codspeed prioritizes test reproducibility, over "maximum performance": so it uses a [simulated cpu instrument](https://codspeed.io/docs/instruments/cpu/overview) to actually count instruction cycles (rather than walltime).  I think it's very likely that 500 nodes will also catch indicate regressions and improvements just as well as 2000 nodes did.  but wanted to point that out

tlambert03 (commented):
all set for review!  let me know if you have questions

dashboard is here:
https://codspeed.io/live-image-tracking-tools/geff

tests/test_bench.py
msschwartz21: Pulling the node count explicitly was a good move. It was something on my mental radar but obviously hadn't gotten around to it 😂

msschwartz21 (approved):
LGTM! Thank you!

Issue #244, opened 2025-07-25
Add properties metadata to the GeffMetadata specification


lxenard:
I would put the props_metadata at the same level as sphere, ellipsoid etc. and make it mandatory even if most of the fields inside are optional.

Candidate fields I have (mandatories in bold):
- **identifier/key**
- name
- (shortname?)
- description
- **dtype**
- unit
- dimension
- **ptype** => property type: node, edge, lineage or tracklet

For now I see 2 ways to do this:
- we keep it simple: dict[identifier, dict[field_name, value]]
- we reproduce the structure for the properties values: a table of ids that would list the identifiers, and an array for each metadata field aligned on the ids

Do we want unique identifiers across all properties? Across ptype?
=> the answer will drive how to store the metadata: all together or separated by ptype

lxenard:
Ah yes of course for the BaseModel class!

- purpose of name/shortname: so both are human friendly names for the property. TrackMate uses the shortname for display in plots and tables, to avoid names too long like 'Mean directional change rate'. So it's a bit redundant but to have a 100% lossless conversion from TrackMate, we need it. So it depens on how lossless you want it to be?
- after discussion => dtype as precise as possible, extracted from property value array 
- dimension is from TrackMate, again used for plotting, might be useful for other tools. It is not the shape but helps characterize the type of the property (I don't find the word for it). In TrackMate is can be: POSITION, LENGTH, TIME, INTENSITY, ANGLE, VELOCITY, AREA... Needed for 100% lossless conversion because for now we don't have a code that can infer the dimension from the unit.
- ptype: yes, we don't need it then!

lxenard:
So after a discussion with @tinevez , we actually don't need a 'dimension' field. JY will deal with the unit to dimension conversion java side.

PR #253, opened 2025-07-28
Add properties metadata to GeffMetadata specification

lxenard:
# Proposed Change
Contributes towards #244, but additional work is needed to close the issue 

Add the following optional fields to GeffMetadata:
- node_props_metadata
- edge_props_metadata
- tracklet_props_metadata
- lineage_props_metadata

# Types of Changes
- New feature or enhancement

- Specification

# Checklist

- [x] I have updated the GeffMetadata and the json schema using `pixi run update-json` if necessary.

# Further Comments
No tests written for the new classes/functions for now => I will write them in the other PR
But the current tests all pass (except spatial-graph because I don't have the dependency)

### DIFF
--- a/src/geff/metadata_schema.py
+++ b/src/geff/metadata_schema.py
@@ -158,21 +158,28 @@ class PropMetadata(BaseModel):
         return self
 
 
-# class PropsMetadataCollection(BaseModel):
-#     """Collection of property metadata organized by component type."""
+def validate_props_metadata(
+    props_metadata: dict[str, PropMetadata],
+    c_type: Literal["node", "edge", "tracklet", "lineage"],
+) -> None:
+    """Check that the keys in the property metadata dictionary match the identifiers
+    in the PropMetadata objects.
 
-#     nodes: dict[str, PropMetadata] | None = Field(  # or could be Sequence instead of dict
-#         None, description="Node property metadata, keyed by property identifier"
-#     )
-#     edges: dict[str, PropMetadata] | None = Field(
-#         None, description="Edge property metadata, keyed by property identifier"
-#     )
-#     tracklets: dict[str, PropMetadata] | None = Field(
-#         None, description="Tracklet property metadata, keyed by property identifier"
-#     )
-#     lineages: dict[str, PropMetadata] | None = Field(
-#         None, description="Lineage property metadata, keyed by property identifier"
-#     )
+    Args:
+        props_metadata (dict[str, PropMetadata]): The property metadata dictionary
+            where keys are property identifiers and values are PropMetadata objects.
+        c_type (Literal["node", "edge", "tracklet", "lineage"]): The type of component
+            to which the property belongs.
+
+    Raises:
+        ValueError: If the key does not match the identifier.
+    """
+    for key, prop_md in props_metadata.items():
+        if key != prop_md.identifier:
+            raise ValueError(
+                f"{c_type.capitalize()} property key '{key}' does not match "
+                f"identifier {prop_md.identifier}"
+            )
 
 
 class RelatedObject(BaseModel):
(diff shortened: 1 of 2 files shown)


src/geff/metadata_schema.py
```diff
@@ -135,6 +135,53 @@ class DisplayHint(BaseModel):
     )
 
 
+class PropMetadata(BaseModel):
+    """Metadata describing a property in the geff graph."""
+
+    identifier: str
+    dtype: str  # TODO: investigate how other packages deal with data types
+    unit: str
```
msschwartz21, line 143:
```suggestion
    unit: str | None = None
```
I think we need to leave this as optional

src/geff/metadata_schema.py
```diff
@@ -135,6 +135,53 @@ class DisplayHint(BaseModel):
...
+    name: str | None = None
+    shortname: str | None = None  # TODO: do we keep this?
```
msschwartz21, line 145: I would prefer not to have this as it's so close to name or identifier

src/geff/metadata_schema.py
```diff
@@ -135,6 +135,53 @@ class DisplayHint(BaseModel):
...
+    description: str | None = None
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> PropMetadata:
+        if not self.identifier:
+            raise ValueError("Property identifier cannot be an empty string.")
+        if not self.dtype:
+            raise ValueError("Property dtype cannot be an empty string.")
```
msschwartz21, line 153: I think pydantic should be doing these checks for you and you don't need to have them explicitly

src/geff/metadata_schema.py
```diff
@@ -135,6 +135,53 @@ class DisplayHint(BaseModel):
...
+    unit: str
+    name: str | None = None
+    shortname: str | None = None  # TODO: do we keep this?
+    description: str | None = None
+
+    @model_validator(mode="after")
+    def _validate_model(self) -> PropMetadata:
+        if not self.identifier:
+            raise ValueError("Property identifier cannot be an empty string.")
+        if not self.dtype:
```
msschwartz21, line 152: I'd like to see if we can check dtype against valid dtypes, but need to figure out where we can get a definitive list.

src/geff/metadata_schema.py
```diff
@@ -208,6 +255,37 @@ class GeffMetadata(BaseModel):
         "Each axis can additionally optionally define a `unit` key, which should match the valid"
         "OME-Zarr units, and `min` and `max` keys to define the range of the axis.",
     )
+
+    # PROPOSITION 1: separate metadata by component type
```
msschwartz21, line 259:
```suggestion
```

src/geff/metadata_schema.py
```diff
@@ -208,6 +255,37 @@ class GeffMetadata(BaseModel):
...
+            "and the values are PropMetadata objects describing the properties."
+        ),
+    )
+    lineage_props_metadata: dict[str, PropMetadata] | None = Field(
+        None,
+        description=(
+            "Metadata for lineage properties. The keys are the property identifiers, "
+            "and the values are PropMetadata objects describing the properties."
+        ),
+    )
```
msschwartz21, line 287: Let's leave these out for now since we haven't formalized the tracklet and lineage property data structure yet. We can follow this pattern when that's ready to go in. 
```suggestion
    tracklet_props_metadata: dict[str, PropMetadata] | None = Field(
        None,
        description=(
            "Metadata for tracklet properties. The keys are the property identifiers, "
            "and the values are PropMetadata objects describing the properties."
        ),
    )
    lineage_props_metadata: dict[str, PropMetadata] | None = Field(
        None,
        description=(
            "Metadata for lineage properties. The keys are the property identifiers, "
            "and the values are PropMetadata objects describing the properties."
        ),
    )
```

msschwartz21 (commented):
Other than a few minor comments this looks good to me! 

@ksugar this is what we were thinking for storing property dtypes. Is there any other metadata you think would be helpful, e.g. a boolean flag for varlen properties?

This PR doesn't handle actually setting these metadata values during writing, but we'll merge that separately so you can have this information if you need it for your PR.

src/geff/metadata_schema.py
lxenard: So let's remove it for now, and I will see if JY can manage this on the java side.
If several tools request it, we can always put it back

src/geff/metadata_schema.py
lxenard: Yes! I also want to check with numpy and co how they deal with is. But it's going to take a bit of time so I think just a string with no checks is fine for now?
I can include this in the next PR

ksugar:
@msschwartz21 After the discussion today, I have a feeling that it is helpful to have an optional metadata that describes how they are encoded. Is unit a mandatory field? If we store strings here, I was wondering what would be the unit.

References:
- https://docs.h5py.org/en/latest/strings.html#how-to-store-raw-binary-data
- https://numpy.org/neps/nep-0055-string_dtype.html
There is also a related discussion here.
https://github.com/saalfeldlab/n5-zarr/issues/74

msschwartz21 (approved):
🎉

PR #255, opened 2025-07-28
Ensure schema is updated in tests, rather than in pre-commit & ci

tlambert03:
Currently, in order to make sure that `geff-schema.json` is up to date with the schema.  the `Check json schema` step runs on CI, and simply fails if the file needs updating.  There is also a `pre-commit` local hook. but because it requires a fully competent environment with geff installed, it only works via pixi/uv, and must be hidden from pre-commit.ci using the "manual" hook stage; as a result, most people will never use it unless they explicitly opt in to manual hook stages.

This PR dispenses with both of those strategies and just checks that the schema is up to date in tests itself.  If the schema is not up to date, tests will fail with a message that you should run `pytest` with the  `--update-schema` flag to allow pytest to update the schema file in place (this is to avoid side-effects without explicit opt-in).  If you then run `pytest --update-schema` it will update the file in place if needed, *and emit a warning that the changes need to be checked in*.  because warnings are errors, that will still be a failure on CI.  However, locally, the user just needs to check in the changes and then tests will pass again.


tlambert03:
all set @msschwartz21

msschwartz21 (approved):

PR #258, opened 2025-07-28
changes np.bool to np.bool_, to work with any numpy version

TeunHuijben:
removing `np.bool`, to make sure geff works with any numpy version

### DIFF
--- a/src/geff/geff_reader.py
+++ b/src/geff/geff_reader.py
@@ -119,7 +119,6 @@ class GeffReader:
 
     def build(
         self,
-        node_mask: NDArray[np.bool_] | None = None,
         node_mask: NDArray[bool] | None = None,
         edge_mask: NDArray[bool] | None = None,
     ) -> InMemoryGeff:
--- a/src/geff/typing.py
+++ b/src/geff/typing.py
@@ -1,6 +1,5 @@
 from typing import Any, TypedDict
 
-import numpy as np
 import zarr
 from numpy.typing import NDArray
 from typing_extensions import NotRequired


tlambert03:
I do think this is a good change... a couple additional thoughts:

a) I'm a bit surprised this wasn't caught by https://github.com/live-image-tracking-tools/geff/pull/250 ... which is now testing against both the min and max versions of stated numpy support.  (what version of python/numpy did you catch it on?)
b) since these are all type hints... they really needn't be evaluated at runtime at all.  if these modules had `from __future__ import annotations__ at the top, it wouldn't have been an error

TeunHuijben:
Hi @tlambert03, thank you for taking a look!

The reason for this change is that I have another project/environment that uses `geff`. Because of other constraints, that environment has `numpy==1.26.4`, which complains about `np.bool` being deprecated.

tlambert03:
yep sounds good.  but let's just make it `bool` rather than `np.bool_` ... since `np.bool_` is simply an alias for the builtin book

tlambert03:
aha!  and I just realized we don't list numpy at all in our dependencies.  that's a problem, and that's why #250 didn't catch this

pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "zarr>=2.18,<4",
     "pydantic>=2.9",
     "networkx>=3.2.1",
+    "numpy",
```
tlambert03, line 19:
```suggestion
    "numpy>=1.26",
    "numpy>=2.1; python_version >= '3.13'",
```

it's a little annoying... but this will fix tests and ensure that users always get a version of numpy for which a wheel has been buillt

src/geff/geff_reader.py
```diff
@@ -119,8 +119,8 @@ def read_edge_props(self, names: list[str] | None = None):
 
     def build(
         self,
-        node_mask: NDArray[np.bool] | None = None,
-        edge_mask: NDArray[np.bool] | None = None,
+        node_mask: NDArray[np.bool_] | None = None,
+        edge_mask: NDArray[np.bool_] | None = None,
```
tlambert03, line 123:
```suggestion
        node_mask: NDArray[bool] | None = None,
        edge_mask: NDArray[bool] | None = None,
```

src/geff/typing.py
```diff
@@ -29,7 +29,7 @@ class PropDictNpArray(TypedDict):
     """
 
     values: NDArray[Any]
-    missing: NotRequired[NDArray[np.bool]]
+    missing: NotRequired[NDArray[np.bool_]]
```
tlambert03, line 32:
```suggestion
    missing: NotRequired[NDArray[bool]]
```

tlambert03:
> I can replace np.bool_ with bool, which passes my tests locally, but pylance doesn't like it, because it is within an NDarray. Any preference?

as long as the current pre-commit passes... go with it.  there are a *ton* of things here that pyright/pylance would complain about, but we're not yet dealing with them

PR #263, opened 2025-07-29
Zarr group to storelike

TeunHuijben:
# Proposed Change
Replaced all (now really) instances of `zarr.group`, we ALWAYS use `StoreLike` for any reading/writing/info/validate etc. etc.

# Types of Changes
- Tests

- typing

# Checklist

### DIFF
--- a/src/geff/networkx/io.py
+++ b/src/geff/networkx/io.py
@@ -88,6 +88,7 @@ def write_nx(
         node_props,
         edge_props,
         axis_names,
+        zarr_format=zarr_format,
     )
 
     # write metadata


src/geff/networkx/io.py
```diff
@@ -74,8 +73,6 @@ def write_nx(
             Defaults to 2.
     """
 
-    group = setup_zarr_group(store, zarr_format)
```
msschwartz21, line 77: Does removing this mean we lose the ability do default to zarr 2?

msschwartz21: I think I fixed this with a commit that I just pushed. It passes the zarr format into write_dicts and then its set correctly there.

msschwartz21 (approved):
LGTM! Thank you for squashing 🐛

PR #264, opened 2025-07-29
remove pixi, add justfile

### LINKED ISSUE #261: Retire pixi (redefine tasks)
@tlambert03, since #250, pixi doesn't work anymore, because it cannot handle the double definitions of packages (like we did for `numcodecs` and `scikit-image` for different Python versions). 

Retiring `pixi` is completely fine with me, I was just wondering how we handle the `pixi.tasks` that are defined in the `pyproject.toml`? Can we run the "pixi" tasks using uv, or should we redefine them as pixi tasks? Since pixi doesn't work anymore, we can just as well remove all pixi references. Apart from the tasks, does removing the pixi-stuff have any other implications?

tlambert03:
closes #261 

@msschwartz21, have a look and let me know how this feels for you.  Quickstart is:

```sh
brew install just
just --list
```

then start playing with commands.

One of the main difference from the pixi task runner is that you don't have the intrinsic ability here to associate a task with a specific environment... but uv makes it easy to ensure that task deps are included by group.   I didn't do it here, but if *don't* want the `uv run` to update the environment in `.venv` you can always add `uv run --isolated ...` to force it to use an ephemeral task-specific env


TeunHuijben:
Awesome, thanks @tlambert03!!!

PR #267, opened 2025-07-29
Big typing cleanup, enable strict typing

tlambert03:
This PR fixes all of the typing issues I could find.  It enables strict mypy type checking, adds deps to pre-commit.  It also makes sure it works with pyright (which catches some things better than mypy, like unbound variables, of which there were a few).

closes #226

I know there's a lot here, so I don't expect this to merge soon, and I can keep it up to date as other PRs get merged.  I'm also sympathetic if you don't want to "impose" this sort of thing on contributors (since there *will* likely be questions about why a given type is failing from time to time).  But... I do think there were many latent bugs that were caught here.

One other thing this does is establish a `_path` module where string literals like `"nodes"`, `"edges"` etc can live... (makes it clearer what the actual spec is)


src/geff/interops/ctc.py
```diff
@@ -70,15 +70,11 @@ def from_ctc_to_geff(
...
-    tracks_file_found = False
-
     for tracks_file in ["man_track.txt", "res_track.txt"]:
         tracks_file_path = ctc_path / tracks_file
         if tracks_file_path.exists():
-            tracks_file_found = True
             break
-
-    if not tracks_file_found:
+    else:
```
tlambert03, line 77: otherwise `tracks_file_path` is potentially unbound

src/geff/interops/trackmate_xml.py
```diff
@@ -465,6 +465,8 @@ def _parse_model_tag(
     it = ET.iterparse(xml_path, events=["start", "end"])
     _, root = next(it)  # Saving the root of the tree for later cleaning.
 
+    units: dict[str, str] = {}
+    attrs_md: dict[str, dict[str, str]] = {}
```
tlambert03, line 469: otherwise potentially unbound

src/geff/testing/data.py
```diff
@@ -182,7 +182,7 @@ def create_dummy_graph_props(
     actual_num_edges = min(num_edges, max_possible_edges)
 
     # Create edges ensuring we don't create duplicates
-    edges: list[list[Any]] = []
+    edges_: list[list[Any]] = []
```
tlambert03, line 189: otherwise type conflict below when you recast to np

src/geff/validators/segmentation_validators.py
```diff
@@ -32,17 +34,12 @@ def has_valid_seg_id(
     """
 
     errors: list[str] = []
-    if isinstance(store, str | Path):
-        store = zarr.DirectoryStore(str(store))
```
tlambert03, line 36: zarr.DirectoryStore doesn't exist in zarr v3 anyway... but was untested

src/geff/metadata_schema.py
```diff
@@ -235,7 +242,7 @@ class GeffMetadata(BaseModel):
     )
 
     geff_version: str = Field(
-        ...,
+        default=GEFF_VERSION,
```
tlambert03, line 245: see `formatted_schema_json` below as a new approach to declare in the schema that geff_version is required.
This does otherwise pose typing problems for anyone trying to create a GeffMetadata without passing version.  (we were lying about the default in order to satisfy the schema)

src/geff/units.py
```diff
@@ -1,10 +1,14 @@
...
+
 import numpy as np
+import numpy.typing as npt
 
 # OME-NGFF 0.5 units
 # https://github.com/ome/ngff/blob/7ac3430c74a66e5bcf53e41c429143172d68c0a4/index.bs#L240-L245
 
-VALID_SPACE_UNITS = [
-    None,
+SpaceUnits: TypeAlias = Literal[
```
tlambert03, line 11: these *could* be used now as pydantic field annotations... but I stopped short of that because it looks like we still want to allow invalid strings and just warn rather than error?

src/geff/utils.py
```diff
@@ -51,90 +53,135 @@ def remove_tilde(store: StoreLike) -> StoreLike:
...
+def validate(store: StoreLike) -> None:
+    """Ensure that the structure of the zarr conforms to geff specification
 
     Args:
         store (str | Path | zarr store): Check the geff zarr, either str/Path/store
 
     Raises:
-        AssertionError: If geff specs are violated
-        ValueError: If store is not a valid zarr store or path doesn't exist
+        ValueError: If geff specs are violated
```
tlambert03, line 63: this might need some discussion.  I'm not a fan of using `assert X` for user-facing errors.  I strictly use them for "things that should never happen".  The main counter argument I've heard is "but I like the one-liner"... which doesn't persuade me :)  but happy to discuss and revert if you feel strongly about this

otherwise: main changes here are just structural.  reusing some logic with some helper functions

src/geff/utils.py
```diff
@@ -157,6 +204,7 @@ def nx_is_equal(g1: nx.Graph, g2: nx.Graph) -> bool:
     nodes_default = len(nodes_attr) * [0]
     nm = iso.categorical_node_match(nodes_attr, nodes_default)
 
+    same_nodes = same_edges = False
```
tlambert03, line 207: otherwise potentially unbound

tests/test_agnostic/test_utils.py
```diff
@@ -36,34 +42,34 @@ def test_validate(tmp_path):
...
     # No nodes
-    with pytest.raises(AssertionError, match="graph group must contain a nodes group"):
+    with pytest.raises(ValueError, match="'graph' group must contain a group named 'nodes'"):
         validate(zpath)
-    z.create_group("nodes")
+    nodes = z.create_group("nodes")
 
     # Nodes missing ids
-    with pytest.raises(AssertionError, match="nodes group must contain an ids array"):
+    with pytest.raises(ValueError, match="'nodes' group must contain an 'ids' array"):
```
tlambert03, line 50: in many of these changes, we're just putting the name of the group in quotes (makes it stand out)

tlambert03 (commented):
just some notes

src/geff/utils.py
melisande-c: At some point I was considering proposing a custom `GeffValidationError` exception, because there may be cases where you'd want to specifically catch it

tlambert03:
just pinging this PR now that the hackathon is over.  I will continue to merge in conflicts as they come, but if you see an opportunity to get this in, that would be great.  Then I think the next "big one" that should happen when you have a down moment is a big public/private refactor

msschwartz21:
I should have some time to go through this PR tomorrow so that we can get in in before the big refactor/regor

msschwartz21 (approved):
Thank you, Talley!!! Everything you did makes sense to me. I left a couple comments/questions, but I think we should merge as soon as the merge conflicts are resolved.

src/geff/geff_reader.py
```diff
@@ -104,23 +119,27 @@ def read_edge_props(self, names: list[str] | None = None):
...
+            if _path.MISSING in prop_group.keys():
+                missing = utils.expect_array(prop_group, _path.MISSING, "edge")
+                prop_dict[_path.MISSING] = missing
             self.edge_props[name] = prop_dict
 
     def build(
         self,
-        node_mask: NDArray[bool] | None = None,
-        edge_mask: NDArray[bool] | None = None,
+        node_mask: NDArray[np.bool_] | None = None,
```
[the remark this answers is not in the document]
tlambert03: that was @TeunHuijben in #258 ... and he changed it to `np.bool_`.  I asked him to make it plain `bool` since I thought it was an alias, which he politely did, but i was wrong :) ... and as he said: pyright complains. 
This works for all versions of numpy (and we know that for sure now since #250 tests both the min and maximum version of stated dependencies across all python versions)

src/geff/metadata_schema.py
```diff
@@ -160,50 +171,48 @@ class PropMetadata(BaseModel):
...
+        default=None,
         description=("Optional description of the property."),
     )
 
-    @model_validator(mode="after")
-    def _validate_model(self) -> PropMetadata:
-        if not self.identifier:
-            raise ValueError("Property identifier cannot be an empty string.")
-        if not self.dtype:
-            raise ValueError("Property dtype cannot be an empty string.")
```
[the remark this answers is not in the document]
tlambert03: yeah: in general I would encourage you to move as many validations into the fields themselves (And *not* in model_validator ... which should be used exclusively for checks that need to access multiple fields).  prefer using things from annotated_types if you can (for basic constraints) or more complex validators if not.  Using annotated_types, for example, ensures that the json schema *also* contains those constraints, as can be seen in the update to the schema here

src/geff/utils.py
tlambert03: yeah: I really don't think `assert something` should ever really be used for a user-facing exception.  as the [python docs say](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement):

> Assert statements are a convenient way to insert debugging assertions into a program:

it's really intended for debugging, usually for something that *should never happen*.  I'm sure there is precedent out there using other ways, but outside of pytest, the primary argument i've heard is just aesthetic: people like the one-liner.  but it's not what assert was meant for, and as you've probably found elsewhere, ruff will also call you out on it if you [add S101](https://docs.astral.sh/ruff/rules/assert/)

src/geff/interops/trackmate_xml.py
```diff
@@ -875,7 +877,7 @@ def _build_geff_metadata(
         node_props_metadata=props_metadata["node_props_metadata"],
         edge_props_metadata=props_metadata["edge_props_metadata"],
         track_node_props={"lineage": "TRACK_ID"},
-        related_objects=[{"type": "image", "path": img_path}] if img_path is not None else None,
+        related_objects=[RelatedObject(type="image", path=img_path)] if img_path else None,
```
tlambert03, line 880: one more comment here. avoid leaning on pydantic's ability to cast dictionaries to model instances for internal code.  Prefer using the actual objects (it makes refactoring much easier)

msschwartz21:
@tlambert03 I'm ready to merge if you're happy!

tlambert03:
Yep go for it!

And also, just to say up front: if you later want to disable either the strict mypy check or the pyright test, that's totally fine :). Or feel free to ping me anytime if one is failing in a confusing way

PR #269, opened 2025-07-30
Update affine model to store numpy array directly, and provide schema hints

### LINKED ISSUE #254: fix: extra should be Mapping, not Any, Matrix should be numpy, not any
the `extra` field is an important place for anyone to put extra information.  As such it really *must* be a `MutableMapping` (not `Any`).  If one person sets it to something like `True`, as done in the tests, then no one else can use it.  Basically, nothing on the model should ever be `Any`.

similarly, the `Affine.matrix` should not be `Any`, it really *should* be a numpy array (otherwise users of the field really can't make any assumptions about how to use it).  This one is a bit harder in pydantic, so I'll open a PR to fix that soon.

tlambert03:
This PR updates the Affine model

closes #254 (in conjunction with #267)

The primary goal here is to:
1. provide better indication of what the matrix must be in the json schema (i.e. list of list of float), which is done via the `json_schema_input_schema` in the TFormMeta object.
2. get rid of `arbitrary_types_allowed` and store an actual numpy array by giving pydantic all the info it needs to cast any object to a well-formed square homogenous matrix (with `Annotated[np.ndarray, TFormMeta]`)

cc @ilan-theodoro please review :)

-----

I stopped there, however I think it's worth discussing (in another PR) whether there should be an `Affine` class *at all*.  One could simply make the GeffMetadata.affine field *directly* a `Annotated[np.ndarray, TFormMeta]`... in which case the schema would look like:

```json
"affine": {
  "anyOf": [
    {
      "items": {
        "items": {
          "type": "number"
        },
        "type": "array"
      },
      "type": "array"
    },
    {
      "type": "null"
    }
  ],
  "default": null,
  "description": "Affine transformation matrix to transform the graph coordinates to the physical coordinates. The matrix must have the same number of dimensions as the number of axes in the graph."
},
```

This does mean you take some of the convenience methods off the `Affine`... but I generally favor pure models that just say what they are, without trying to add too much of a "convenience API" on top of it (as tempting as it is! 😄 )... since it's just all the more to possibly break.  It doesn't stop you from having a functional module that adds some of these conveniences, but it does makes the model clean and direct.  The main reason to *keep* the full `Affine` object is if you intend to eventually have something *other* than one single `matrix` field on it

### DIFF
--- a/src/geff/affine.py
+++ b/src/geff/affine.py
@@ -3,13 +3,16 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Annotated, Any
 
 import numpy as np
-from pydantic import BaseModel, Field, GetCoreSchemaHandler, model_validator
+from pydantic import BaseModel, Field, GetCoreSchemaHandler, model_validator, version
 from pydantic_core import core_schema
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
     from numpy.typing import DTypeLike, NDArray
+
+
+PYDANTIC_211 = tuple(int(x) for x in version.VERSION.split(".")[:2]) >= (2, 11)
 
 
 def _validate_tform(val: Any) -> np.ndarray:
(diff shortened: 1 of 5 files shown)


src/geff/affine.py
```diff
@@ -24,55 +74,31 @@ class Affine(BaseModel):
...
-
-    matrix: Any = Field(
-        ..., description="Homogeneous transformation matrix as list of lists (ndim+1, ndim+1)"
+    matrix: Annotated[np.ndarray, TFormMeta] = Field(
+        ...,
+        description="Homogeneous transformation matrix (ndim+1, ndim+1)",
     )
 
-    @model_validator(mode="after")
-    def validate_matrix(self) -> Affine:
```
tlambert03, line 37: side comment, for the future:   Try not to use `model_validator` in cases where you only want to validate a single field.  Use `field_validator` for that (or, in complex cases, like here, you can [customize validation with `__get_pydantic_core_schema__`](https://docs.pydantic.dev/latest/concepts/types/#as-an-annotation))

PR #273, opened 2025-07-30
feat(read): Support rustworkx and spatial-graph backends in unified read function

### LINKED ISSUE #260: Use test_read function to test all the in-memory implementations (or kick the code)
Might interact strangely with the optional dependencies. Needs investigation

melisande-c:
# Proposed Change

- Add `RUSTWORKX` and `SPATIAL_GRAPH` to `SupportedBackend` enum.
- Update `construct_rx` to take array arguments not a single dict of arrays.
- Ensure `construct_rx` and `construct_sg` are only imported into `read.py` if the relevant backends have been installed. If someone tries to use the `read` function with a backend that hasn't been installed an error will be raised.
- Small update to `read` function signature: `backend_kwargs` are now real python kwargs not a dict. This allows the overloads for each backend to include the additional kwargs, e.g. spatial-graph overload has the `position_attr`  argument.
- Updated all the test helper functions for `test_read` for rustworkx and spatial graph. Now all backends are tested identically.
  - For spatial graph there is an additional `sg_get_node_spatial_props` function because all the spatial props are combined into one position prop, whereas the other backends (nx and rx) keep the different spatial props (e.g. `"x"`, `"y"`, `"t"`) separate.

N.b. currently the whole of `test_io.py` will be skipped if any of the backends are not installed because of the `pytests.importskip` at the top of the file. There may be a better way to do this but as long as one of the CIs runs with everything installed I think it should be fine.

Resolves #260.

# Types of Changes
- New feature or enhancement

- Read implementation

# Checklist

- [x] I wrote tests for the new implementation using standard fixtures supplied in conftest.py.
- [x] I updated pyproject.toml with new dependencies if needed.
- [x] I added a function to tests/bench.py to benchmark the new implementation.


src/geff/io/read.py
```diff
@@ -120,8 +111,20 @@ def get_construct_func(backend: SupportedBackend) -> ConstructFunc[Any]:
...
+                    "use `pip install 'geff[rx]'`"
+                )
+            return construct_rx
+        case SupportedBackend.SPATIAL_GRAPH:
+            if construct_sg is None:
+                raise ImportError(
+                    "spatial-graph is not installed. To read a GEFF to the spatial-graph backend "
+                    "please use `pip install 'geff[spatial-graph]'`"
+                )
+            return construct_sg
```
tlambert03, line 127: since both of these `io` modules already raise a similar error, what do you think of just:


```suggestion
        case SupportedBackend.RUSTWORKX:
            from geff.rustworkx.io import construct_rx

            return construct_rx
        case SupportedBackend.SPATIAL_GRAPH:
            from geff.spatial_graph.io import construct_sg

            return construct_sg
```


and them remove the `find_spec` logic above?

src/geff/io/read.py
```diff
@@ -1,22 +1,43 @@
...
+    construct_rx: ConstructFunc | None = rx_io.construct_rx
+else:
+    construct_rx = None
+sg_spec = find_spec("spatial_graph")
+if sg_spec is not None:
+    import geff.spatial_graph.io as sg_io
+
+    construct_sg: ConstructFunc | None = sg_io.construct_sg
+else:
+    construct_sg = None
```
tlambert03, line 41:
```suggestion
```

(see comment below)

src/geff/io/supported_backends.py
```diff
@@ -10,5 +10,5 @@ class SupportedBackend(str, Enum):
     """
 
     NETWORKX = "networkx"
-    # GRAPH_DICT can be removed when another backend is added, it is currently needed for overloads
-    GRAPH_DICT = "graph_dict"
+    RUSTWORKX = "rustworkx"
+    SPATIAL_GRAPH = "spatial_graph"
```
tlambert03, line 14:
```suggestion
    SPATIAL_GRAPH = "spatial-graph"
```

(suggest using package name rather than top level module)

src/geff/io/read.py
```diff
@@ -157,9 +170,10 @@ def read(
     # using Literal because mypy can't seem to narrow the enum type when chaining functions
     backend: Literal[
         SupportedBackend.NETWORKX,
-        SupportedBackend.GRAPH_DICT,
+        SupportedBackend.RUSTWORKX,
```
tlambert03, line 173: I kinda wonder whether the Enum is helping here or hurting...

One implication here is that you expect people to have to import `geff.SupportedBacked` to select a backend (rather than just `geff.read(..., backend='rustworkx')`).  Also, as you noticed, it does seem to make the type narrowing more challenging.  Here would be another option:


```python
SupportedBackend = Literal["networkx", "rustworkx", "spatial-graph"]


@overload
def get_construct_func(
    backend: Literal["networkx"],
) -> ConstructFunc[nx.Graph | nx.DiGraph]: ...
@overload
def get_construct_func(
    backend: Literal["rustworkx"],
) -> ConstructFunc[rx.PyGraph | rx.PyDiGraph]: ...
@overload
def get_construct_func(
    backend: Literal["spatial-graph"],
) -> ConstructFunc[sg.SpatialGraph | sg.SpatialDiGraph]: ...
def get_construct_func(backend: SupportedBackend) -> ConstructFunc[Any]: ...

@overload
def read(
    store: StoreLike,
    validate: bool = ...,
    node_props: list[str] | None = ...,
    edge_props: list[str] | None = ...,
    *,
    backend: Literal["networkx"],
) -> tuple[nx.Graph | nx.DiGraph, GeffMetadata]: ...
@overload
def read(
    store: StoreLike,
    validate: bool = ...,
    node_props: list[str] | None = ...,
    edge_props: list[str] | None = ...,
    *,
    backend: Literal["rustworkx"],
) -> tuple[rx.PyGraph | rx.PyDiGraph, GeffMetadata]: ...
@overload
def read(
    store: StoreLike,
    validate: bool = ...,
    node_props: list[str] | None = ...,
    edge_props: list[str] | None = ...,
    *,
    backend: Literal["spatial-graph"] = ...,
    position_attr: str = "position",
) -> tuple[sg.SpatialGraph | sg.SpatialDiGraph, GeffMetadata]: ...
def read(
    store: StoreLike,
    validate: bool = True,
    node_props: list[str] | None = None,
    edge_props: list[str] | None = None,
    *,
    backend: SupportedBackend = "networkx",
    **backend_kwargs: Any,
) -> tuple[Any, GeffMetadata]: ...
```

it still provides all the useful type checking and IDE autocompletion, doesn't require an extra import, and doesn't require explicit listing of each backend.

Other things to note here:
- use `= ...` for defaults in overloads (type checkers don't pay attention to them anyway, and it's one more less thing to have to keep in sync)
- i suggest making backend a keyword only argument.  not only does it help with type checking, it makes it easier to modify the signature later, and also reads better when calling the function

if you definitely do prefer and want to stick with the Enum, that's fine too, I would just tell mypy to hush up with:

```python
construct_func = get_construct_func(backend)  # type: ignore[call-overload]
```

src/geff/io/read.py
```diff
@@ -103,8 +88,14 @@ def get_construct_func(
...
+) -> ConstructFunc[rx.PyGraph | rx.PyDiGraph]: ...
+
+
+@overload
+def get_construct_func(
+    backend: Literal[SupportedBackend.SPATIAL_GRAPH],
+) -> ConstructFunc[sg.SpatialGraph | sg.SpatialDiGraph]: ...
 
 
 def get_construct_func(backend: SupportedBackend) -> ConstructFunc[Any]:
```
tlambert03, line 101: we could choose to just define a backend as "a module/class with a construct() function" ... this whole thing could be reduced to something like:

```
getattr(geff, backend_obj).construct
```

with no need to duplicate all the overloads.  As long as it's not a public-facing function, that's plenty stable and testable without boilerplate?

(that might need to wait for a follow up PR though, the entirety of the geff isn't public 😄 )

tests/test_io.py
```diff
@@ -34,21 +42,41 @@ def is_expected_type(graph, backend: SupportedBackend):
...
+        axes = metadata.axes
+        if axes is None:
+            raise ValueError("No axes found for spatial props")
+        axes_names = [ax.name for ax in axes]
+        if name in axes_names:
+            return sg_get_node_spatial_props(graph, name, nodes, axes_names)
+        else:
+            # TODO: is this the best way to access node attributes?
+            return getattr(graph.node_attrs[nodes], name)
     else:
```
tlambert03, line 80: these things all feel like further support for the concept of a backend namespace. 

So: rather than having the `read.py` do all of the conditional logic to say "if backend == X then ...", maybe have a full backend protocol, where each backend knows how to convert itself to standard types you will ask for elswhere:

```python
class NxBackend(Backend):
    construct = construct_nx
    get_nodes = nx.Graph.nodes
    get_edges = nx.Graph.edges
    get_node_prop = ...
```

?

src/geff/io/supported_backends.py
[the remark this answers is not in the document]
melisande-c: I could do, but I never access the access the value attribute because the enum inherits from string you can do `SupportedBackend.NETWORKX == "networkx"` and it will return True, someone can input the string into a function instead of a `SupportedBackend` object and it will still work

src/geff/io/read.py
melisande-c: I did consider this but I didn't want to add import time as an overhead to the read func, but tbh I don't know anything about how python imports work under the hood and if they are cached, if you don't think it's a problem then yeah it looks neater

src/geff/io/read.py
melisande-c: Ah I didn't know you could use `...` for defaults in overloads, thanks!!!

src/geff/io/read.py
tlambert03: But... to be clear.  

These are all just thoughts and you can merge this now if you want or if you disagree :) 

it's already a nice improvement!

src/geff/io/supported_backends.py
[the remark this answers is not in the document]
tlambert03: Yes, the equality will work at runtime, but that's not what the type hint is saying

src/geff/io/read.py
melisande-c: With the Literal it is better, it was so annoying when I realised mypy wasn't working with the chained literals. I feel like I have been hurt with literals before in a code base where I didn't catch a typo somewhere but it is pretty contained here

src/geff/io/read.py
tlambert03: Oh yeah: I'd say that's premature optimization:)

Once imported, a reimport is simply a dictionary lookup.  Way less code, microsecond lookup

tests/test_io.py
melisande-c: yeah... I was avoiding writing a whole backend class since these are only used in the tests but it does make sense

src/geff/io/read.py
tlambert03: And also; wouldn't you rather only pay that import time cost (once) when you're actually using that backend, and not _every_ time you import geff?

src/geff/io/read.py
tlambert03: agree.  I definitely *don't* encourage/suggest you actually use literal strings `"networkx"` all over the codebase.  But as a user-facing parameter, in this particular case, I think it's quite sufficient.  If you find yourself referring to the backend in many different places, then I would agree with some constant somewhere (either enum of something like [this](https://github.com/live-image-tracking-tools/geff/pull/267/files#diff-2471405108377e79371ae9ba59ffeec6ffda3bfb8b496c3a4127de4b6847118a)

src/geff/io/read.py
melisande-c: makes sense! I will update

src/geff/io/read.py
melisande-c: ahh ok, ok, just for the function signature makes sense

tests/test_io.py
melisande-c: yeah would be better for maintenance

melisande-c:
@cmalinmayor @tlambert03 I will be on holiday until the 6th of August, I am happy for anyone to push any changes, or I can make changes when I am back, or I can make the changes in a second PR.

tlambert03 (approved):
looks good to me!

src/geff/spatial_graph/io.py
```diff
@@ -168,7 +170,7 @@ def construct_sg(
     node_props: dict[str, PropDictNpArray],
     edge_props: dict[str, PropDictNpArray],
     position_attr: str = "position",
-) -> sg.SpatialGraph:
+) -> sg.SpatialGraph | sg.SpatialDiGraph:
```
tlambert03, line 173: the remaining type error you're getting is indicating that `read_sg` needs to be updated to reflect the new return type of `construct_sg` (which is correct):

```python
-> tuple[sg.SpatialGraph | sg.SpatialDiGraph, GeffMetadata]
```

(because sg.SpatialDiGraph is not a subclass of sg.SpatialGraph here)

PR #284, opened 2025-08-04
Validate units with cf-units package

mkitti:
# Proposed Change
Expands the valid units to include anything that the `cf-units` python package (and thus the [UDUNITS-2 C library](https://www.unidata.ucar.edu/software/udunits/)) recognizes. This relaxes the definition with regard to OME-NGFF in that we allow forms outside of the enumerated strings in the OME-NGFF specification which units SHOULD be.

In particular, this allows for plural forms such as "micrometers" or "seconds".

# Types of Changes
- Bugfix (non-breaking change which fixes an issue)
  - JSON in specification.md is now valid
- New feature or enhancement
  - Expands units to include those recognized by cf-units

Which topics does your change affect? Delete those that do not apply.
- Specification

# Checklist
Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [x] I have read the [developer/contributing](https://github.com/live-image-tracking-tools/geff/blob/main/CONTRIBUTING) docs.
- [x] I have added tests that prove that my feature works in various situations or tests the bugfix (if appropriate).
- [x] I have checked that I maintained or improved code coverage.
- [x] I have written docstrings and checked that they render correctly by looking at the docs preview (link left as a comment on the PR).

## If you changed the specification
- [x] I have checked that any validation functions and tests reflect the changes.
- [x] I have updated the GeffMetadata and the json schema using `pytest --update-schema` if necessary.
- [x] I have updated docs/specification.md to reflect the change.
- [x] I have updated implementations to reflect the change. (This can happen in separate PRs on a feature branch, but must be complete before merging into main.)

# Further comments

We could remove the list of valid units completely if we use this definition.

PR #285, opened 2025-08-05
Check units with pint

mkitti:
# Proposed Change
Check units with [`pint`](https://pint.readthedocs.io/en/stable/index.html)

Supercedes https://github.com/live-image-tracking-tools/geff/pull/284

# Types of Changes
  - specification.md example is now valid
- New feature or enhancement
  - Units are now validated against pint's default unit registry

- Specification

# Checklist

### DIFF
--- a/src/geff/valid_values.py
+++ b/src/geff/valid_values.py
@@ -96,7 +96,13 @@ def validate_space_unit(unit_name: str) -> bool:
         bool: True if a space unit is a KNOWN valid unit.
         False if the unit is not known. The unit may be valid.
     """
-    return unit_name in VALID_SPACE_UNITS or cf_units.Unit(unit_name).convert(1, "micrometer")
+    try:
+        return unit_name in VALID_SPACE_UNITS or cf_units.Unit(unit_name).convert(1, "micrometer")
+    except ValueError as err:
+        # Conversion to micrometer should not error
+        if str(err).startswith("Unable to convert"):
+            return False
+        raise
 
 
 def validate_time_unit(unit_name: str) -> bool:
--- a/src/geff/valid_values.py
+++ b/src/geff/valid_values.py
@@ -1,8 +1,8 @@
 import encodings
 import pkgutil
 
-import cf_units
 import numpy as np
+import pint
 
 # -----------------------------------------------------------------------------
 # Unit validation
@@ -73,6 +73,7 @@ VALID_AXIS_TYPES = [
     "channel",  # TODO: discuss
 ]
 
+ureg = pint.UnitRegistry()
 
 def validate_axis_type(axis_type: str) -> bool:
     """Validate axis type against standard list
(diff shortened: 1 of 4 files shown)


src/geff/valid_values.py
```diff
@@ -95,7 +98,7 @@ def validate_space_unit(unit_name: str) -> bool:
         bool: True if a space unit is a KNOWN valid unit.
         False if the unit is not known. The unit may be valid.
     """
-    return unit_name in VALID_SPACE_UNITS
+    return unit_name in VALID_SPACE_UNITS or ureg(unit_name).check("[length]")
```
msschwartz21, line 101: Can you expand this docstring (and the one for time) to mention checking against pint?

msschwartz21 (changes requested):
This is a nice use for `pint`! 

Just to recap, the behavior for units now is they must either be listed on the OME Zarr units or be parsed as a standard unit by `pint`. If neither criteria is met, we raise a warning. If the unit is only recognized by pint and not by OME Zarr then we issue a warning. Is that correct?

Can you add a note to this [section](https://github.com/live-image-tracking-tools/geff/blob/main/docs/specification.md#geff-metadata) of the docs about allowing support for any units recognized by `pint`?

tlambert03:
> Actually we always warn if the unit is not one exactly stated in the OME-NGFF specification.

by the way... I also feel like this "warn and leave as is" behavior is probably something that will have to go away in the future.  The whole point of a specification is to give downstream consumers something reliable to use.  So if someone enters an invalid unit, and you *don't* raise an exception, you're just leaving it for *every* downstream application to decide what to do with it, and I suspect the vast majority of them will just discard it.  I think the spec is the _right_ place to ensure that units are  (strictly) one of a valid set.  and raise (not warn) otherwise

mkitti:
At the moment, we do not have a closed set of units.  We have a "SHOULD" suggestion from OME-NGFF, not a "MUST".

The question is thus do we only accept only the OME-NGFF units as I listed previously? Do we accept plural forms of those units?

Could geff be used beyond microscopy tracking?

If the answer is yes to the latter questions, then I think we need pint.

If we want to restrict to only NGFF units, then we should error not warn if another unit is presented.

tlambert03:
I think it should be a MUST (either leave them out or make them useful).  


> Could geff be used beyond microscopy tracking?
> If the answer is yes to the latter questions, then I think we need pint.

I would say the answer would be yes *when* you have an application in mind.  not "just in case"

src/geff/valid_values.py
mkitti: I expanded the doc string in https://github.com/live-image-tracking-tools/geff/pull/285/commits/8452df1670c8e57860a1ded32c5324e73baee238

mkitti:
> For example?

`decaseconds` and `decameters` are missing from the OME-NGFF text.

This [`.zattrs`](https://s3.janelia.org/keller-lab/kittisopikulm/idr0062A/6001240.zarr/.zattrs) refers to "microns" but seems to validate with [ome-ngff-validator](https://ome.github.io/ome-ngff-validator/):
https://ome.github.io/ome-ngff-validator/?source=https://s3.janelia.org/keller-lab/kittisopikulm/idr0062A/6001240.zarr/

As far as I can tell ome-ngff-validator does not seem to validate units at all:
https://github.com/ome/ome-ngff-validator/issues/52

ome-zarr-models does not currently validate any SHOULD statements:
https://github.com/ome-zarr-models/ome-zarr-models-py/issues/170

I will also note that the OME-XML unit abbreviations are not included in this list.

Would you prefer to make pint an optional dependency?

Issue #301, opened 2025-08-12
Make PropMetadata Mandatory


cmalinmayor:
This will require: updating writer implementation to write it, updating the reader implementations to use it, updating the specification, writing tests.

PR #302, opened 2025-08-12
279 update property metadata

### LINKED ISSUE #279: Update metadata specification
More precisely:
- yeet 'encoding' => we agreed that only utf8 string encoding is supported so no need to keep this field
- add a 'varlength' dtype => discussion with @cmalinmayor: we need to store the info that a prop is variable length somewhere, and storing it in its own field seems overkill. More intuitive to store it as a dtype and should integrate well

cmalinmayor:
# Proposed Change
Closes #279 as discussed with @lxenard. Needed for variable length properties implementation. I also moved PropMetadata to it's own file because _schema.py was getting very long and confusing.

Next: #301 

# Types of Changes
- New feature or enhancement

- Specification
- Metadata

# Checklist


msschwartz21 (approved):
I'm assuming that we have the actual update to the spec describing how the varlength dtype is handled in the next PR. Otherwise docs are fine with these changes.

🚜

PR #292, opened 2025-08-11
API reorganization

msschwartz21:
# Proposed Change
Majorly reorganizes where code lives and what the import patterns are with the goal of making our public facing api docs easy to navigate and read. Closes #183. This will introduce breaking changes for libraries that are depending on geff (sorry!)

Still to do, but will happen in another PR:
- A comprehensive review of how the docstrings are rendering and adding some missing docstrings
- Reorganize tests to match the new module structure

# Types of Changes
What types of changes does your code introduce? Delete those that do not apply.
- Documentation update
- Maintenance (e.g. dependencies, CI, releases, etc.)

Which topics does your change affect? Delete those that do not apply.

# Checklist
Put an x in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code.

- [ ] I have read the [developer/contributing](https://github.com/live-image-tracking-tools/geff/blob/main/CONTRIBUTING) docs.
- [ ] I have added tests that prove that my feature works in various situations or tests the bugfix (if appropriate).
- [ ] I have checked that I maintained or improved code coverage.
- [ ] I have written docstrings and checked that they render correctly by looking at the docs preview (link left as a comment on the PR).

## If you changed the specification
- [ ] I have checked that any validation functions and tests reflect the changes.
- [ ] I have updated the GeffMetadata and the json schema using `pytest --update-schema` if necessary.
- [ ] I have updated docs/specification.md to reflect the change.
- [ ] I have updated implementations to reflect the change. (This can happen in separate PRs on a feature branch, but must be complete before merging into main.)

## If you have added or changed an implementation
- [ ] I wrote tests for the new implementation using standard fixtures supplied in conftest.py.
- [ ] I updated pyproject.toml with new dependencies if needed.
- [ ] I added a function to tests/bench.py to benchmark the new implementation.

# Further Comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #308, opened 2025-08-20
Add backend protocol

melisande-c:
# Proposed Change
In #273 it was suggested a backend protocol should be defined to remove a few if-else blocks for when different backends are used.

This PR adds a `Backend` `Protocol` class that can be used to directly type module instance objects. It enforces that a backend module should have the functions `construct`, `get_node_ids`, `get_edge_ids`, `get_node_prop` and `get_edge_prop` plus the constant `GRAPH_TYPES`.

The `construct_<backend>` functions have all been renamed to `construct` and all the other required functions (e.g. `get_node_ids`) have been added to the modules, the functionality was previously defined in `tests/test_graph_libs/test_api_wrapper.py`.

`test_api_wrapper.py` has been refactored to use the `get_node_ids` etc. functions in the backend modules.

The `README.md` in `_graph_libs` has been updated to reflect the changes. 

# Types of Changes
- New feature or enhancement

# Checklist

# Further Comments


msschwartz21 (commented):
These changes generally make sense to me and do seem to make the code simpler. However, I'm not super familiar with python protocols so I would appreciate it if @tlambert03 has a chance to take a look this PR.

msschwartz21:
@melisande-c I was looking more closely at how the backend protocol works especially with regards to the construct function. If we add read to the backend protocol, would it eliminate the need for the read function and associated overloads in `_api_wrapper`?

src/geff/_graph_libs/_backend_protocol.py
```diff
@@ -0,0 +1,125 @@
...
+                with warning.
+            *args (Any): Additional positional arguments used for construction.
+            **kwargs (Any): Additional keyword arguments used for construction.
+
+        Returns:
+            graph (T): The graph object.
+        """
+        ...
+
+    def get_node_ids(self, graph: T) -> Sequence[Any]:
```
tlambert03, line 59: another thought I'm having... I believe these methods `get_x` are largely there for tests right? (and were partially why we went down this path of a backend protocol at all correct)?  

I'm wondering if, since they really are used largely as a convenience for testing (and *feel* different than the `construct/read` functions, perhaps this should be two different "protocols", where all these methods that take a backend graph object live on something like a:

```
class BackendGraphAdapter(Generic[T]):
    def __init__(self, graph):
        self.graph = graph
    def get_node_ids(self) -> Sequence[Any]: ...
    def get_edge_ids(self) -> Sequence[tuple[Any, Any]]: ...
```

and then in tests, or any place you want a "uniform" api... rather than

```python
backend_module = get_backend(backend)
backend_module.get_node_ids(graph)
backend_module.get_edge_ids(graph)
```

it would be
```python
adapter = get_adaptor(graph)
adapter.get_node_ids()
adapter.get_edge_ids()
```

melisande-c:
> @melisande-c I was looking more closely at how the backend protocol works especially with regards to the construct function. If we add read to the backend protocol, would it eliminate the need for the read function and associated overloads in _api_wrapper?

@msschwartz21 We could indeed add the read function to the `Backend` class (and if we subclass it, as @tlambert03  suggests, then it wouldn't have to reimplement it for every backend). It depends whether you want users to have access to a read function in the geff top level API, without dealing with the backends or people should use the `get_backend` function.

i.e.

```python
from geff import read

graph = read(source) # networkx is default backend

# or for other backends
graph = read(source, backend="rustworkx")
```
vs 
```python
from geff import get_backend

backend = get_backend() # networkx is default
graph = backend.read(source)

# or for other backends
backend = get_backend("rustworkx")
graph = backend.read(source)
```

---

@tlambert03 

> The one thing I'm playing with in my head is **whether it might be beneficial actually subclass the protocol in the backends.?**

The reason why I decided to keep the backends as modules originally was because I started working on this before the the big reorganisation in #292. I thought a valid way users might interact with a particular backend would be to directly import the module so then if there was a class in there it might be a bit awkward to use. But it seems after the reorganisation this isn't an intended way to use different backends so I think having IDE feedback and linting would be a nice benefit.

> i.e. take advantage of the fact that `Protocol` is an `abc.ABC`

huh, I didn't realise this, just had a look and it's the `typing._ProtocolMeta` that inherits from `abc.ABCMeta`, pretty cool. But I think we'll be fine without the `abstractmethod` decorator because I don't think we need the runtime checks, the type checker can cover it (when inheriting from the protocol).

> I suppose it might feel slightly awkward to have to instantiate the backend (but if you really hated that, you can make them all `@staticmethods`

I don't think instantiating the class in the `get_backend` function is a big deal. But I might still make the construct function a `staticmethod` to be explicit that it shouldn't depend on an internal state of the class.

I'm happy to make these changes, I can do it by the end of the week.

melisande-c:
I have now added tests for the unified write function and I discovered some bugs in the spatial graph `write` function, both of which I have fixed.

1) When accessing the spatial graph node and edge properties the node order has to be specified by doing `graph.node_attrs[graph.nodes`, if you just do `graph.node_attrs` the properties can end up in the wrong order which is what was happening.

2) Secondly the write function was assuming the axis order was always ["t", "z", "y", "x"] and if there were 3 dimensions then it would always be ["z", "y", "x"] but this was not consistent with the construct function which allows for example  ["t", "y", "x"] and also for the axes to be in a different order. I fixed this by requiring axis_names in the write function either by specifying `axis_name` or by having them present in the `GeffMetadata`.

msschwartz21 (commented):
I'm really liking this new structure! I think it's a lot cleaner and going to be easier to maintain and test in the long run. I didn't see any issues in my initial pass of the code, but I'd like to get a couple sets of eyes on this before we merge it in. 

I had to introduce a new pin to fix the docs build on another PR https://github.com/live-image-tracking-tools/geff/pull/322/commits/65f445de258757b9671dc3db5cdb9cf4705f85d5. If #310 isn't merged first, you can cherry pick that change over to this branch.

msschwartz21 (commented):
I went to update the code blocks in #317 and it reminded me that I think we should just write a universal construct function as well to go along with read/write. Advanced users who want to use the GeffReader directly in order to filter nodes/edges will then need the construct function. It looks nicer to have the universal function instead of grabbing it from the backends.

src/geff/__init__.py
```diff
@@ -1,44 +1,19 @@
...
     "GeffReader",
     "read",
-    "read_nx",
-    "read_rx",
-    "read_sg",
     "validate_structure",
-    "write_nx",
-    "write_rx",
-    "write_sg",
+    "write",
```
msschwartz21, line 18: Can we also expose `construct` here?

melisande-c:
Oh also codecov is counting the `...` in the empty function bodies of the protocols as uncovered. Is there a config somewhere to ignore lines with only ellipses?

melisande-c:
> > I updated the benchmark tests to use the new `get_backend` function for retrieving the respective read and write functions but now CodSpeed doesn't know which tests are which. What should I do? I can make the tests take a callable as a parametrised argument again or is there a way to fix it in CodSpeed?
> 
> I wonder if adding an `id` to parametrize would fix the problem as long as the id matches the name that CodSpeed had before?

Oh good idea, I'll try it

src/geff/_graph_libs/_api_wrapper.py
```diff
@@ -18,137 +16,125 @@
...
-            **kwargs (Any): Optional kwargs for constructing the `in_memory_geff`.
-
-        Returns:
-            A graph object instance for a particular backend.
-        """
-        ...
+    NxGraph: TypeAlias = nx.Graph | nx.DiGraph
+    RxGraph: TypeAlias = rx.PyGraph | rx.PyDiGraph
+    SgGraph: TypeAlias = sg.SpatialGraph | sg.SpatialDiGraph
+    SupportedGraphType: TypeAlias = NxGraph | RxGraph | SgGraph
```
tlambert03, line 24: just noting ... (but i don't think anything needs to be done here... this PR has been open long enough and is a very nice improvement already):

the `get_backend_from_graph_type`, is the only thing that requires these SupportedGraphTypes, and is the only remaining thing in this `_api_wrapper` that requires backend-specific knowledge (i.e. where the abstraction leaks a bit). 

One could have a `def supports()` method on the backends, so rather than *you* doing the loop with 

```
    for backend, graph_type in BACKEND_TYPE_MAP.items():
        if isinstance(graph, graph_type):
            return get_backend(backend)
```

it would be `for backend in backends: if backend.supports(graph)`...  and then all the `BACKEND_TYPE_MAP` and other logic gets removed here.

src/geff/_graph_libs/_api_wrapper.py
```diff
@@ -18,137 +16,125 @@
...
 
 
+# This function is used to create BACKEND_TYPE_MAP after the get_backend definition
+# Only installed backends will be added to the backend type map
+def _create_backend_type_map() -> dict[SupportedBackend, tuple[type[SupportedGraphType], ...]]:
+    mapping: dict[SupportedBackend, tuple[type[SupportedGraphType], ...]] = {}
+    backends: tuple[SupportedBackend] = get_args(SupportedBackend)
+    for backend in backends:
+        if find_spec(MODULE_NAMES[backend]) is not None:
+            backend_io = get_backend(backend)
```
tlambert03, line 77: also just noting, but not necessary to change:

this func gets called on module import, and actually calls `get_backend` for every backend... so we are greedily importing *all* third party things (networkx, rustworkx, spatial-graph)  – assuming they are there – even if the end user is never going to use them.  Which incurs a possibly unnecessary import time cost.

Also  `find_spec(MODULE_NAMES[backend])`, is sort of duplicated/detached logic from simply `try: get_backend(backend)`.  It doesn't actually save you an import (since you're about to do it anyway), and *might* diverge in the future from the actual import requirements of the specific backend

src/geff/_graph_libs/_api_wrapper.py
```diff
@@ -172,11 +158,171 @@ def read(
...
+@overload
+def write(
+    graph: NxGraph,
+    store: StoreLike,
+    metadata: GeffMetadata | None = ...,
+    axis_names: list[str] | None = ...,
+    axis_units: list[str | None] | None = ...,
+    axis_types: list[str | None] | None = ...,
+    zarr_format: Literal[2, 3] = ...,
+) -> None: ...
```
tlambert03, line 228: these overloads on `write` are a bit unnecessary... the return type doesn't change based on the input `graph`,  so a simple `Union` is sufficient.  Yes, it does allow you to specifically hint at a rx-specific `node_id_dict` ... but honestly that backend-specific argument is a bit unfortunate anyway, and should perhaps just be added as a general argument (even if it's not used by all backends)

src/geff/_graph_libs/_backend_protocol.py
```diff
@@ -0,0 +1,150 @@
...
+            node_props (list of str, optional): The names of the node properties to load,
+                if None all properties will be loaded, defaults to None.
+            edge_props (list of str, optional): The names of the edge properties to load,
+                if None all properties will be loaded, defaults to None.
+            data_validation (ValidationConfig, optional): Optional configuration for which
+                optional types of data to validate. Each option defaults to False.
+            *args (Any): Additional args that may be accepted by the backend when reading the
+                data.
+            **kwargs (Any): Additional kwargs that may be accepted by the backend when reading the
+                data.
```
tlambert03, line 90: somewhat unfortunate to have to duplicate all of these docstrings between both the backend classes, and the public api functions.  If we commit to making the backend objects just internal implementation details, you could consider deleting these arguments (so they don't have to be manually kept in sync with the function docstrings) and simply leave a note to backend-creating developers to see the docstrings for the corresponding functions in `_api_wrapper`?

(but, just keeping them in sync is also fine)

tlambert03 (approved):
i think this looks awesome!  Left some minor thoughts, but LGTM

src/geff/_graph_libs/_api_wrapper.py
melisande-c: hmm yeah I see that now, it would probably be best to just try to import all the backends in the _api_wrapper module, not the function. I will update it

src/geff/_graph_libs/_api_wrapper.py
melisande-c: yeah I originally just had the rustworkx one for the extra node_id_dict, I don't know why I added the other overloads in, probably just trying to satisfy my need for symmetry aha

PR #315, opened 2025-08-25
Add note to specification that group can use .geff

### LINKED ISSUE #126: Naming convention: Name a geff zarr group with .geff suffix
Right now, when inspecting some top-level zarr group on disk, to understand that there is a zarr group within it that contains a geff graph, I need to open the `top-level.zarr/my-graph-name/.zattrs` file.

It would be nice for humans, as well as readers, to understand from the folder name, that it contains a geff graph.
Zarr does something similar, by having a naming convention of `.zarr` for the top-level group. 

@DragaDoncila and I would suggest to recommend naming the zarr group (=directory) that contains the geff graph (see [spec](https://github.com/live-image-tracking-tools/geff/blob/main/docs/specification.md)) with a `.geff` suffix, and that the default geff writer would follow this naming convention.

In the above example, this would correspond to `top-level.zarr/my-graph-name.geff`.

msschwartz21:
# Proposed Change
Closes #126 

# Types of Changes
- Documentation update

- Specification

# Checklist


cmalinmayor (approved):
🍅

PR #320, opened 2025-08-28
Enforce node ids must be uint

### LINKED ISSUE #176: Node IDs should be unsigned integers and this should be validated
Potentially we provide a utility that converts node IDs to integers and then keeps the original string IDs as a property.

msschwartz21:
Closes #176 

The main issue was that when `write_dicts` constructs an array of node_ids the step of `np.asarray` doesn't guarantee a uint output even if we started with uint node ids.

# Types of Changes
- Tests

- Specification
- Core io

# Checklist

### DIFF
--- a/src/geff/core_io/_base_write.py
+++ b/src/geff/core_io/_base_write.py
@@ -60,8 +60,16 @@ def write_dicts(
 
     if len(node_ids) > 0:
         nodes_arr = np.asarray(node_ids)
+        # Check if we can cast to uint
+        if any(nodes_arr < 0):
+            raise ValueError("Cannot write a geff with node ids that are negative")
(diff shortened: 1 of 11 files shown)


src/geff/core_io/_base_write.py
```diff
@@ -60,8 +60,16 @@ def write_dicts(
 
     if len(node_ids) > 0:
         nodes_arr = np.asarray(node_ids)
+        # Check if we can cast to uint
+        if any(nodes_arr < 0):
+            raise ValueError("Cannot write a geff with node ids that are negative")
+        if not np.issubdtype(nodes_arr.dtype, np.integer):
```
cmalinmayor, line 66: If we validate the uniqueness of the node ids, it should be after this cast.

PR #322, opened 2025-08-31
Update segmentation validation functions to use InMemoryGeff

### LINKED ISSUE #310: Refactor segmentation validation functions to use a seg_id array directly instead of loading it from a store
This will make those functions easier to test and maintain

msschwartz21:
Closes #310 

@AnniekStok let me know if this pattern works for your downstream uses cases? Having it run on a memory geff makes these functions much easier to test and maintain than going directly from the store, but I don't want to create too much friction. I also considered trying to include these function in the new framework for [data validation](https://github.com/live-image-tracking-tools/geff/blob/4f66537a143c5cb94ca058e272b35948cd80f040/src/geff/validate/data.py#L31), but that requires having the segmentation already loaded which is a bit different from the current pattern. 

# Types of Changes

- Validate

# Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ requires-python = ">=3.10"
 license = { text = "MIT License" }
 dynamic = ['version']
 dependencies = [
-    "typer>=0.14.0,<0.17.0",
+    "typer>=0.14.0",
     "zarr>=2.18,<4",
     "pydantic>=2.11",
     "networkx>=3.2.1",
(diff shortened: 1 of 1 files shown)


AnniekStok (approved):
Thanks @msschwartz21 ! Yes, I think this works. @cmalinmayor, I created a separate funtracks branch (https://github.com/funkelab/funtracks/pull/82) to test using the InMemoryGeff instead of the StoreLike, it seems to work on my side, what do you think?

PR #325, opened 2025-09-05
Switch to running benchmarking on codspeed macro to collect walltime

msschwartz21:
I'm testing if recording walltime during codspeed benchmarking will be more informative and potentially stabilize our benchmark times. I'm still not entirely clear on what causes all of the variability in the spatial-graph benchmarks and so further investigation may be needed.

# Types of Changes

# Checklist


msschwartz21:
Stick with normal "not walltime" benchmarking and maybe just skip spatial graph benchmarking for now

PR #326, opened 2025-09-05
[WIP] Monorepo

tlambert03:
this is a *super* work-in-progress (more of a proof of principle really) showing what a mono-repo would look like with 2 packages (`geff-spec`, and `geff`).

it's currently broken: i moved metadata schema into geff-spec but haven't update geff.

cc @msschwartz21


msschwartz21:
> If i understand correctly... it's the same reason as they are currently required inside of src: namely: that is the top level module folder (the thing you import). Which is different from the package (packages/name) folder (the thing with the pyproject.toml file)

Sorry that's my bad. You're totally right and somehow I convinced myself that it was a change from how we had it before even though it's absolutely not 😂

msschwartz21:
Closing b/c now replaced by #329

PR #342, opened 2025-09-26
Validate when writing geffs

### LINKED ISSUE #333: Validate before writing
Currently it's possible to write an invalid geff that won't pass validation upon reading. We should run validation on the InMemoryGeff before calling write_arrays.

msschwartz21:
# Proposed Change
Closes #333

I considered a few different approaches to how to make sure we didn't write an invalid geff. A major constraint is that our structural validation function takes a StoreLike as input which means that the geff has to be written before we can validate the structure. Given this constraint, I introduced a utility function that deletes a geff that we have written while doing its best to handle the different possible types in StoreLike and avoiding delete any members of the zarr that aren't geff members.

Alternatively, I considered generalizing the `validate_structure` function to operate on either a StoreLike or an InMemoryGeff but the approaches that I tried were leading to a ton of code duplication and maintenance overhead. I think the deleting option is going to be easier to maintain in the long run even if its slightly less elegant and a bit inefficient.

# Types of Changes

- Core io
- Validate

# Checklist


msschwartz21:
@melisande-c we're hitting some weird issues with the `get_backend_from_graph_type` function all of a sudden ([ex1](https://github.com/live-image-tracking-tools/geff/actions/runs/18045369623/job/51354741268?pr=342), [ex2](https://github.com/live-image-tracking-tools/geff/actions/runs/18045369623/job/51354741411?pr=342)). I shouldn't have changed anything in this PR that would cause an issue as far as I know. Would you mind taking a look and seeing if you have any ideas about what might be happening?

msschwartz21:
@melisande-c I've narrowed the bug down a bit. It looks like when all of those tests are failing `AVAILABLE_BACKENDS` is an empty list which I presume means that `_import_available_backends` is not being called prior to trying to use the `write` function. I don't have a good hypothesis for why this is happening now or why it only shows up on certain configurations of OS/dependencies.

melisande-c:
> @melisande-c I've narrowed the bug down a bit. It looks like when all of those tests are failing `AVAILABLE_BACKENDS` is an empty list which I presume means that `_import_available_backends` is not being called prior to trying to use the `write` function. I don't have a good hypothesis for why this is happening now or why it only shows up on certain configurations of OS/dependencies.

Hmm that is very strange, especially as it was working before, I will try to take a look when I have a spare moment

packages/geff/src/geff/core_io/_base_write.py
```diff
@@ -247,6 +262,13 @@ def write_arrays(
         metadata = compute_and_add_axis_min_max(metadata, node_props)
     metadata.write(geff_store)
 
+    if structure_validation:
+        try:
+            validate_structure(geff_store)
+        except ValueError as e:
+            _delete_invalid_geff(geff_store)
```
cmalinmayor, line 269: Need to pass the zarr_format here

cmalinmayor (changes requested):
Just very minor suggestions - the passing of the zarr format is the only one that definitely needs to be fixed, the others just might be nice.

PR #348, opened 2025-09-30
345 axis offset

### LINKED ISSUE #345: Missing `offset` geff metadata
Hi folks,

I'm very thankful for the recent release and significant codebase improvements in the latest version of `geff`.

However, a bit too late, I noticed the removal of the Affine method #331, which we were already using.

I agree that `scale` and `offset` would address our use case, but `offset` is missing.

Should I open a PR to fix this?

Moreover, I think not having `Affine` is detrimental to the spec in the long term.
It's more general than `offset` and `scale`, and it will likely be requested in the future.

msschwartz21:
# Proposed Change
Closes #345 by adding axis offset to the spec

Inheriting from #346 

# Types of Changes
- New feature or enhancement

- Specification
- Metadata

# Checklist

### DIFF
--- a/geff-schema.json
+++ b/geff-schema.json
@@ -212,6 +212,19 @@
           "default": null,
           "description": "Optional, the unit after applying the `scale` value to the data. If `scaled_unit` is set, a `scale` value must also be provided.",
           "title": "Scaled Unit"
+        },
+        "offset": {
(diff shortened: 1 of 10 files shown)


JoOkuma:
Thanks a lot @msschwartz21

PR #352, opened 2025-10-01
Always use geff-spec version in metadata (not geff)

msschwartz21:
# Proposed Change
I noticed after using the ctc converter that we were still ending up with the geff version in the metadata instead of the geff-spec version. This PR fixes that bug in one of our metadata utility functions.

# Types of Changes

- Metadata

# Checklist


cmalinmayor (approved):
🌮

PR #353, opened 2025-10-02
Add `MissingDependencyError` for filtering from other `ImportErrors`

melisande-c:
# Proposed Change
Add a custom `MissingDependencyError` that will be raised if one tries to import a module that uses custom backends when the relevant package is not installed. Many libraries call their error for missing optional dependencies the same thing.

The reason to do this is that currently we might accidentally swallow an import error when it is not actually related to missing dependencies and it might be hard to debug. This happened in #342. This way we can filter for this error separately.

Additionally added a couple of typing fixes.

# Types of Changes
- New feature or enhancement

# Checklist


msschwartz21 (approved):
Thank you for this fix and figuring out the bug in #342! I was totally stumped!

Issue #337, opened 2025-09-23
Improve CI runtime and caching efficiency


msschwartz21:
Currently tests can take up to 40 minutes to run. For example, [this run](https://github.com/live-image-tracking-tools/geff/actions/runs/17950528340) took a total of 51 minutes with the longest individual test taking 36 minutes. 

- [x] Improve total runtime for tests
- [ ] Improve caching behavior

PR #354, opened 2025-10-03
Remove and consolidate duplicate test cases to speed up testing time

### LINKED ISSUE #330: Testing API wrapper code ignores missing values
https://github.com/live-image-tracking-tools/geff/blob/1a0dbb37769d4d48c248e4a142ac3dd52845598e/tests/test_graph_libs/test_api_wrapper.py#L88-L93

@melisande-c How would you suggest we adapt this to not assert that the missing values are the same? The specific problem I am having right now is that the [GraphAdapter](https://github.com/live-image-tracking-tools/geff/blob/1a0dbb37769d4d48c248e4a142ac3dd52845598e/src/geff/_graph_libs/_graph_adapter.py#L32) doesn't really support missing values (the networkx implementation actually gives a KeyError if one of the values is missing, which is unfortunate). Should `get_node_prop` return a PropDictNpArray with values and missing? And then the testing helper can only compare the non-missing values? For the networkx case, we basically need to then fill in the missing values with something dtype and shape compatible... which is now feeling very reminiscent of the dict writer.

Alternatively, we can change the GraphAdapter to have a more individual node/dge API (e.g. `has_node_prop(node, prop)` and `get_node_prop(node, prop)`. Then the testing has to loop over each node/edge and check individually that either it's missing or the value is correct. This feels more in line with testing that the graph backend implementations are correct.

I haven't finished thinking this through, but I wanted to document the problem and see if you or anyone else (@msschwartz21 @tlambert03 ?) had thoughts.

msschwartz21:
# Proposed Change
Starts to address #337 and #188 by refactor our tests for `geff._graph_libs`
- Removes direct tests of different backends and instead tests all backends through api_wrapper tests
- Eliminates some parametrization by consolidating many different data types into a single test case with multiple properties instead of testing each data type individually

I should try to fix #330 in this PR, but haven't gotten to it yet. 

# Types of Changes
- Tests

# Checklist

### DIFF
--- a/packages/geff/tests/test_graph_libs/test_api_wrapper.py
+++ b/packages/geff/tests/test_graph_libs/test_api_wrapper.py
@@ -20,8 +20,6 @@ sg = pytest.importorskip("spatial_graph")
 
 
 # TODO: missing test cases
-# - empty graph
-# - missing properties
 # - different store types
 
 
@@ -112,6 +110,7 @@ class Test_api_wrapper:
             include_y=include_spatial,
             include_z=include_spatial,
             include_varlength=backend != "spatial-graph",
+            include_missing=backend != "spatial-graph",
         )
 
         graph, metadata = read(store, backend=backend)
@@ -146,6 +145,7 @@ class Test_api_wrapper:
             include_y=include_spatial,
             include_z=include_spatial,
             include_varlength=backend != "spatial-graph",
+            include_missing=backend != "spatial-graph",
         )
 
         in_memory_geff = read_to_memory(store)
@@ -182,6 +182,7 @@ class Test_api_wrapper:
             include_y=include_spatial,
             include_z=include_spatial,
             include_varlength=backend != "spatial-graph",
+            include_missing=backend != "spatial-graph",
         )
 
         # this will create a graph instance of the backend type
--- a/packages/geff/tests/test_graph_libs/test_api_wrapper.py
+++ b/packages/geff/tests/test_graph_libs/test_api_wrapper.py
@@ -68,51 +68,43 @@ def _assert_graph_equal_to_geff(
                     assert expected_val == actual_val
 
 
-@pytest.mark.parametrize("node_id_dtype", ["uint8", "uint16"])
-@pytest.mark.parametrize(
-    "node_axis_dtypes",
-    [
-        {"position": "double", "time": "double"},
-        {"position": "int", "time": "int"},
-    ],
-)
-@pytest.mark.parametrize(
-    "extra_edge_props",
-    [
-        {"score": "float64", "color": "uint8"},
-        {"score": "float32", "color": "int16"},
-        {},
-    ],
-)
-@pytest.mark.parametrize("directed", [True, False])
-@pytest.mark.parametrize("include_t", [True, False])
-@pytest.mark.parametrize("include_z", [True, False])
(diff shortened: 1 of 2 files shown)


packages/geff/tests/test_graph_libs/test_api_wrapper.py
```diff
@@ -79,130 +74,248 @@ def _assert_graph_equal_to_geff(
...
-) -> None:
-    backend_module: Backend = get_backend(backend)
-
-    store, memory_geff = create_mock_geff(
+        directed,
+        include_t,
+        include_spatial,
+        backend,
+    ) -> None:
+        # if include_spatial is False and backend == "spatial-graph":
```
cmalinmayor, line 131: I assume delete this?

packages/geff/tests/test_graph_libs/test_api_wrapper.py
```diff
@@ -18,16 +20,9 @@
...
-    {"position": "int", "time": "int"},
-]
-extra_edge_props = [
-    {"score": "float64", "color": "uint8"},
-    {"score": "float32", "color": "int16"},
-    {},
-]
+
+# TODO: missing test cases
+# - different store types
```
cmalinmayor, line 25: Claude says you tested this actually 🤷 if you didn't maybe make it clearer whats still missing

packages/geff/tests/test_graph_libs/test_api_wrapper.py
```diff
@@ -79,130 +74,248 @@ def _assert_graph_equal_to_geff(
...
+            original_graph = create_graph(
+                ndims=1,
+                node_dtype="uint64",
+                node_attr_dtypes={"pos": "float32[1]"},
+                edge_attr_dtypes={},
+                position_attr="pos",
+            )
+        else:
+            raise NotImplementedError(
+                f"Backend {backend} not tested in Test_empgy_graph.test_write"
```
cmalinmayor, line 309:
```suggestion
                f"Backend {backend} not tested in Test_empty_graph.test_write"
```

packages/geff/tests/test_graph_libs/test_api_wrapper.py
```diff
@@ -79,130 +74,248 @@ def _assert_graph_equal_to_geff(
...
+        graph_adapter = NxBackend.graph_adapter(graph)
+
+        _assert_graph_equal_to_geff(graph_adapter, memory_geff)
+        assert metadata.extra["foo"] == "bar"
+        assert metadata.extra["bar"]["baz"] == "qux"
+
+
+@pytest.mark.parametrize("backend", get_args(SupportedBackend))
+class Test_api_wrapper_simple:  # tests that only need backend parametrization
+    def test_write_axis_lists_override_metadata(self, tmp_path, backend):
```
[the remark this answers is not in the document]
cmalinmayor: Yes, SG should check if the number of axes doesn't match the position shape and raise an error

cmalinmayor (changes requested):
🍅 Again, minor things only! You can ping me again after the spatial graph update and then I'll approve.

packages/geff/tests/test_graph_libs/test_api_wrapper.py
msschwartz21: Hahah thanks Claude :sweat-smile: forgot to delete the todo

cmalinmayor (approved):
🎺 🧌

PR #364, opened 2025-10-21
Eliminate overwrite behavior and instead check for existing geff before writing

### LINKED ISSUE #361: Review and update "overwrite" behavior
Currently we have overwrite options in the converters and that option may or may not exist in the core write functions. Regardless, I don't think we've tested it and it's not clear what the behavior should be.

Current opinion based on offline discussion with @cmalinmayor is that write functions should check if a geff already exists and throw an error if that is the case. We can direct users to the delete geff function and let them handle the "overwrite" themselves if that is the behavior they want.

msschwartz21:
# Proposed Change
Closes #361. If overwrite is True, deletes the geff and writes a new geff. Exposed in `geff.write`, all convert functions and `write_arrays`.

# Types of Changes

- Convert
- Core io

# Checklist

### DIFF
--- a/packages/geff/tests/test_core_io/test_base_write.py
+++ b/packages/geff/tests/test_core_io/test_base_write.py
@@ -196,6 +196,7 @@ class TestWriteArrays:
         ):
             write_arrays(store, **new_geff, overwrite=True)
             check_equiv_geff(store, new_store)
+
     def test_write_props_metadata(self):
         _, memory_geff = create_simple_3d_geff()
 


packages/geff/tests/test_core_io/test_core_utils.py
```diff
@@ -139,3 +141,30 @@ def test_mem_store(self):
             delete_geff(store)
             root = zarr.open(store)
             assert "geff" not in root.attrs
+
+
+class Test_check_for_geff:
+    def test_path(self, tmp_path):
+        geff_path = tmp_path / "test.geff"
+        # DONE
```
cmalinmayor: DONE?

msschwartz21:
I added overwrite behavior to our very top level and very bottom level functions. This means that there is not an overwrite option in backend.write functions, but I think that's ok since they aren't publicly exposed.

cmalinmayor (approved):
👍 🦖 💕

Issue #223, opened 2025-07-23
TrackMate XML to GEFF doesn't support segmentation


lxenard:
The segmentation associated with each node is variable in length and cannot be added as a regular property in the GEFF. Need support for polygons.

Note: The XML parser is currently capable of extracting the segmentation data but is not passing it to GEFF on purpose to avoid crashes.

PR #366, opened 2025-10-22
Add storage of spot ROI coordinates as a varlength property

lxenard:
# Proposed Change
Now the TrackMate converter extracts TrackMate spots ROI coordinates and stores them in the geff graph as a variable length property. Solves #223.

# Types of Changes
- New feature or enhancement

- Convert

# Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -131,7 +131,7 @@ jobs:
       contents: write # this permission is required for generate_release_notes
     steps:
       - name: Download built artifact to dist/
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: Packagesgeff
           path: dist
@@ -172,7 +172,7 @@ jobs:
       contents: write # this permission is required for generate_release_notes
     steps:
       - name: Download built artifact to dist/
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: Packagesgeff-spec
           path: dist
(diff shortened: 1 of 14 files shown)


lxenard:
The heavy lifting was already done so it was quite straightforward to do!
But I have an issue that I believe is not related to my code. But I can be wrong so I still want to check with you. The metadata I write in the geff has filled fields (dtype, name, unit), but if I read back the geff with read() these fields are replaced by default values... (float64, None). Any ideas?

lxenard:
So I've added some tests:
- check that ROI data end up in the GEFF, in `test_from_trackmate_xml_to_geff`
- check that PropMetadata are correctly built -> `test_build_geff_metadata`
- check that the metadata is still correct after writing to GEFF and reading back -> `test_metadata_info_stability`
The last one FAILS, node_props_metadata end up with default values. I think the issue is in the read() function 🤔

lxenard:
> Thank you for the additional tests! I used the failing case to identify that it's actually an issue with the write function. We do some amount of automatic populating of the props metadata and are currently failing to preserve any existing metadata. Either I or Caroline should be able to tackle a bug fix for that sometime next week.

Ah great! I did take a quick look at the read() and could not identify the issue, but it didn't cross my mind that write() could be the culprit 😂

lxenard:
> The properties metadata generated exclusively from looking at the xml features contains two additional props `MANUAL_SPOT_COLOR` and `SHAPE_INDEX` which are not included after writing. I'm guessing this is because those features aren't actually present on the graph so they get lost somewhere along the way.

Yes, in the example XML, these 2 props are defined in the metadata but not actually present in the data. The converter detects that and removes them from the metadata in `_check_component_props_consistency().

> The props metadata generated at write time contains three additional properties ID, TRACK_ID and name which I'm guessing are creating while constructing the graph and then picked up at write time.

For these props, it's the exact inverse from above: in the XML, those are props present in  the data but absent from the metadata. The converter builds the missing metadata with default values (at least I think, I need to check to be 100% sure).

@msschwartz21 I will make a full final check of all of this tomorrow and let you know.

PR #369, opened 2025-10-30
Preserve existing PropMetadata when writing props arrays

msschwartz21:
# Proposed Change
Fixes the bug found in #366 where parts of property metadata were lost during write if they were set by the user prior to write.

# Types of Changes

- Metadata

# Checklist

### DIFF
--- a/packages/geff-spec/src/geff_spec/utils.py
+++ b/packages/geff-spec/src/geff_spec/utils.py
@@ -154,7 +154,8 @@ def add_or_update_props_metadata(
         GeffMetadata object with updated props metadata.
 
     Warning:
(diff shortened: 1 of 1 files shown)


packages/geff/tests/test_graph_libs/test_api_wrapper.py
```diff
@@ -4,6 +4,7 @@
 import numpy as np
 import pytest
 import zarr
+import zarr.storage
```
cmalinmayor: Was this just a bug?

cmalinmayor (approved):
🍅 Thanks :)

PR #407, opened 2026-03-31
Dataframe/csv to geff

### LINKED ISSUE #319: CSV to geff converter
@AnniekStok has really nice configurable code for this in funtracks. @msschwartz21 I can propose moving that here in a separate PR, to see if it changes your mind if you don't have to write it 😆

_Originally posted by @cmalinmayor in https://github.com/live-image-tracking-tools/geff/pull/185#discussion_r2301145498_

cmalinmayor:
# Proposed Change
Convert a dataframe to a GEFF (and thus a csv read with pandas)
Closes #319 
@AnniekStok This would hopefully take over the functionality of your funtracks PR https://github.com/funkelab/funtracks/pull/186 - we can just call dataframes_to_memory_geff after splitting the edges into their own df. Current implementation doesn't support list elements, but I think we should just write them into separate columns (unsquish on write) and then manipulate the InMemoryGeff after calling dataframes_to_memory_geff to resquish them as needed. Just need a way to save which ones should be squished, which can go in custom metadata.

Open questions: Should I add a benchmark? Should I implement the squishing? If so, how should I let the user tell us which columns to squish on read - optional argument of dict from new name to old column names, save GeffMetadata with csvs and add that dict to the extras, both as options?

# Types of Changes
- New feature or enhancement

- Convert
- Core io

# Checklist

# Further Comments
I would love at least the default value refactoring to live in GEFF, along with the construct_props helper. If we keep the dataframe specific code in funtracks, at least we can avoid duplicating constructing the values and missing arrays with the correct dtype of default values.

### DIFF
--- a/packages/geff/src/geff/convert/_dataframe.py
+++ b/packages/geff/src/geff/convert/_dataframe.py
@@ -27,7 +27,7 @@ def geff_to_dataframes(
     """
     Convert a GEFF store to a pandas DataFrame.
 
-    Properties with more than 2 dimensions cannot be converted and will be skipped
+    Properties with more than 2 dimensions cannot be converted and will be skipped.
     Properties with two dimensions where the second dimension is > 1 will be unpacked
     into separate columns with the name "{prop_name}_{dim_index}"
 
@@ -92,9 +92,10 @@ def geff_to_dataframes(
 def geff_to_csv(store: StoreLike, outpath: Path | str, overwrite: bool = False) -> None:
     """Convert a geff store to two csvs of nodes and edges
 
-    Properties with more than 2 dimensions cannot be exported and will be skipped
+    Properties with more than 2 dimensions cannot be exported and will be skipped.
     Properties with two dimensions where the second dimension is > 1 will be unpacked
-    into separate columns with the name "{prop_name}_{dim_index}"
+    into separate columns with the name "{prop_name}_{dim_index}", and saved in extra
+    metadata.
 
     Args:
         store (StoreLike): Path to store or StoreLike object
@@ -139,9 +140,27 @@ def dataframes_to_memory_geff(
         edge_source_col: Name of the source node column in edge_df. Defaults to "source".
         edge_target_col: Name of the target node column in edge_df. Defaults to "target".
 
+    Raises:
+        ValueError: If node_df is missing the node_id_col column, or edge_df is
+            missing the edge_source_col or edge_target_col columns.
+
     Returns:
         InMemoryGeff: A dict with metadata, node_ids, edge_ids, node_props, and edge_props.
     """
+    # Validate required columns
+    if node_id_col not in node_df.columns:
+        raise ValueError(
+            f"node_df must contain a {node_id_col!r} column. Found columns: {list(node_df.columns)}"
+        )
+    missing_edge_cols = [
+        col for col in (edge_source_col, edge_target_col) if col not in edge_df.columns
+    ]
+    if missing_edge_cols:
+        raise ValueError(
+            f"edge_df must contain {missing_edge_cols} column(s). "
+            f"Found columns: {list(edge_df.columns)}"
+        )
+
     # Extract node IDs
     if len(node_df) > 0:
         node_ids = np.asarray(node_df[node_id_col]).astype(np.uint64)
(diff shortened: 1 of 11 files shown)


msschwartz21 (commented):
- [x] The "Conversion to GEFF" documentation page needs to be updated with a note about this new functionality

Re benchmarking: my first inclination is yes, but I just checked and we're not benchmarking any of the other converters. So maybe we just open an issue for future conversion benchmarking.

Need to make a more careful pass, but wanted to just initially flag these documentation concerns.

packages/geff/src/geff/core_io/_utils.py
```diff
@@ -232,6 +232,90 @@ def _get_common_type_dims(arr_seq: Sequence[ArrayLike | None]) -> tuple[np.dtype
...
+        return type(value)(0)
+    elif isinstance(value, bool):
+        return False
+    elif isinstance(value, int | float):
+        return 0
+    elif isinstance(value, str):
+        return ""
+    else:
+        return value
+
```
AnniekStok: Yes this works for our motile-tracker geff import/export :)

packages/geff/src/geff/core_io/_utils.py
```diff
@@ -232,6 +232,90 @@ def _get_common_type_dims(arr_seq: Sequence[ArrayLike | None]) -> tuple[np.dtype
...
+        value: A non-None example value to determine the default from.
+
+    Returns:
+        A default value with the same type (and shape, for the fallback case).
+    """
+    if isinstance(value, np.generic):
+        return type(value)(0)
+    elif isinstance(value, bool):
+        return False
+    elif isinstance(value, int | float):
```
msschwartz21: Preserve python dtypes too

cmalinmayor:
@msschwartz21 this is ready except the pandas version issue that I am still debugging 🙄

packages/geff/src/geff/convert/_dataframe.py
```diff
@@ -103,3 +113,239 @@ def geff_to_csv(store: StoreLike, outpath: Path | str, overwrite: bool = False)
     mode = "w" if overwrite else "x"
     node_df.to_csv(node_path, mode=mode)
     edge_df.to_csv(edge_path, mode=mode)
+
+
+def dataframes_to_memory_geff(
```
msschwartz21: I don't totally love having this as a public function since we don't follow the same pattern with the other conversion functions. If it's something where you still want access to the functionality, can we change it to private so it's still available for use, but not explicitly part of our conversion API?

packages/geff/src/geff/convert/_dataframe.py
```diff
@@ -103,3 +113,239 @@ def geff_to_csv(store: StoreLike, outpath: Path | str, overwrite: bool = False)
...
+        )
+
+    # Infer the smallest integer dtype from node IDs (edge IDs are a subset)
+    raw_node_ids = np.asarray(node_df[node_id_col]) if len(node_df) > 0 else np.empty((0,))
+    id_dtype = _infer_int_dtype(raw_node_ids) if len(raw_node_ids) > 0 else np.dtype(np.uint8)
+
+    # Extract node IDs
+    node_ids = raw_node_ids.astype(id_dtype)
+    if len(node_ids) == 0:
+        node_ids = np.empty((0,), dtype=id_dtype)
```
msschwartz21, line 171:
```suggestion
    # Infer the smallest integer dtype from node IDs (edge IDs are a subset)
    if len(node_df) > 0:
        raw_node_ids = np.asarray(node_df[node_id_col])
        id_dtype = _infer_int_dtype(raw_node_ids)
        node_ids = raw_node_ids.astype(id_dtype)
    # Default to uint8 if no nodes 
    else:
        id_dtype = np.uint8
        node_ids = np.empty((0,), dtype=id_dtype)
```

packages/geff/src/geff/core_io/_utils.py
```diff
@@ -232,6 +232,88 @@ def _get_common_type_dims(arr_seq: Sequence[ArrayLike | None]) -> tuple[np.dtype
...
+    Args:
+        value: A non-None example value to determine the default from.
+
+    Returns:
+        A default value with the same type (and shape, for the fallback case).
+    """
+    if isinstance(value, np.generic):
+        return type(value)(0)
+    elif isinstance(value, bool | int | float):
+        return type(value)(0)
```
msschwartz21:
```suggestion
    if isinstance(value, np.generic | bool | int | float):
        return type(value)(0)
```

packages/geff/tests/test_core_io/test_core_utils.py
```diff
@@ -102,6 +132,49 @@ def test_warn_if_writing_zarr_format_3(self, tmp_path: Path):
...
 
 
+class Test_construct_props:
+    @pytest.mark.parametrize(
+        ("values", "expected_values", "expected_missing", "expected_dtype"),
+        [
+            # Python native types
+            ([True, None, False], [True, False, False], [False, True, False], np.bool_),
+            ([1, None, 3], [1, 0, 3], [False, True, False], np.int64),
+            ([1.5, None, 3.5], [1.5, 0, 3.5], [False, True, False], np.float64),
```
msschwartz21: Is this behavior part of what we have had some issues opened about, e.g. data getting upcast to int/float64 when it's not necessary? Maybe outside of the scope of this PR, but potentially we should just do a review of this behavior across the repo

msschwartz21 (changes requested):

msschwartz21 (changes requested):
A couple questions and tweaks before we merge

packages/geff/tests/test_core_io/test_core_utils.py
cmalinmayor: Anywhere we call np.asarray() on an integer thing produces np.<u>int64 dtype. I moved the _infer_integer_dtype() helper function into the utils.py and called it in construct_props to fix as much as possible within the scope of the repo, but we should examine other uses of this pattern.

msschwartz21 (approved):
🚀 🦖

PR #411, opened 2026-04-09
GEFF IO improvement

JoOkuma:
Hi @cmalinmayor 

# Proposed Change

Optimizes zarr I/O in `core_io` by replacing boolean-mask indexing with `oindex` integer indexing for reads, and adding explicit chunking with zarr v3 sharding for writes.

While profiling `tracksdata`, I noticed two performance issues:

1. `_load_prop_to_memory` converts boolean masks to Python lists via `mask.tolist()` before indexing zarr arrays. For sparse selections (e.g. 5 nodes out of 8674) this is ~7x slower than using integer indices with `oindex`, because zarr has to process the entire boolean array instead of just reading the relevant chunks.

2. No explicit chunking when writing arrays — zarr's auto-chunking splits trailing dimensions unnecessarily. For example, `edge_ids` with shape `(189870, 2)` gets chunked as `(47468, 1)`, putting each column in a separate chunk.

Also fixes a bug: the per-node boolean mask was applied to the varlength `DATA` array, which is a flat 1D concatenation with a completely different length. This crashes on zarr v3 with `VindexInvalidSelectionError`. The DATA array must always be loaded in full since deserialization uses byte offsets from VALUES.

### Benchmarks

Dataset: BF-C2DL-HSC/01 (8674 nodes, 189870 edges)

**`reader.build()` selecting 5 nodes:**

| Version | Time |
|---------|------|
| Before (`mask.tolist`) | ~155 ms |
| After (`oindex`) | ~21 ms |

**File count on zarr v3 write:**

| Version | Files on disk |
|---------|--------------|
| Before (auto-chunking) | 119 |
| After (sharding) | 67 |

**Chunking examples:**

| Array | Before | After |
|-------|--------|-------|
| `edge_ids` (189870, 2) uint64 | (47468, **1**) | (189870, **2**) |
| `bbox` (8674, 6) int64 | (4337, 6) | (8674, 6) |
| `mask/data` (4192807,) uint64 | (131026,) | (1048576,) |

# Types of Changes

- New feature or enhancement

Which topics does your change affect?

- Core io

# Checklist

# Further Comments

The change is split into two commits:

1. **Reading**: adds `_load_zarr_subset()` using `oindex` and fixes the varlength DATA bug. The boolean mask API of `_load_prop_to_memory` and `build()` is unchanged — the conversion to integer indices happens internally.

2. **Writing**: adds `_write_zarr_array()` that chunks only along the first dimension (power-of-two size, ~8 MiB target). On zarr v3 it uses sharding so each array becomes a single file with sub-chunks inside. On zarr v2 it just uses the larger explicit chunks. Empty arrays fall back to direct assignment.

All 116 `test_core_io` tests pass.
I haven't added new tests yet — the existing masked-read tests (`test_build_w_masked_nodes`, `test_load_prop_into_memory`, etc.) already exercise the new code paths since the public API is unchanged.
Happy to add targeted benchmarks or edge-case tests if needed.

### DIFF
--- a/packages/geff/src/geff/core_io/_base_write.py
+++ b/packages/geff/src/geff/core_io/_base_write.py
@@ -307,6 +307,73 @@ def write_arrays(
             raise ValueError(e.args[0] + message) from e
 
 
+_TARGET_CHUNK_BYTES = 1 << 23  # 8 MiB — power-of-two, close to 10 MB
+
+
+def _compute_first_dim_chunk(shape: tuple[int, ...], itemsize: int) -> int:
+    """Return a power-of-two first-dimension chunk size targeting ~8 MiB per chunk.
+
+    Trailing dimensions are kept whole so only the first dimension is chunked.
+    """
+    row_bytes = int(np.prod(shape[1:])) * itemsize if len(shape) > 1 else itemsize
+    n_rows = _TARGET_CHUNK_BYTES // max(row_bytes, 1)
+    # Round down to nearest power of two
+    if n_rows >= 1:
+        n_rows = 1 << (n_rows.bit_length() - 1)
+    return max(1, min(shape[0], n_rows))
+
+
+def _write_zarr_array(
+    group: Any,
+    path: str,
+    data: np.ndarray,
+) -> None:
+    """Write a numpy array to a zarr group with sensible chunking.
+
+    Chunks only along the first dimension (trailing dimensions are kept whole)
+    with a power-of-two first-dimension size targeting ~8 MiB per chunk.
+
+    On zarr v3 sharding is used so every array becomes a single file: the shard
+    (outer chunk) spans the entire first dimension while inner sub-chunks are
+    ~8 MiB each.
+
+    Args:
+        group: The zarr group to write into.
+        path: The path within the group.
+        data: The numpy array to write.
+    """
+    import zarr as _zarr
+
+    if data.shape[0] == 0:
+        group[path] = data
+        return
+
+    first_dim = _compute_first_dim_chunk(data.shape, data.dtype.itemsize)
+    chunks = (first_dim, *data.shape[1:])
+
+    use_sharding = (
+        not _zarr.__version__.startswith("2")
+        and hasattr(group, "metadata")
+        and getattr(group.metadata, "zarr_format", 2) == 3
+    )
+
+    if use_sharding:
+        # Shard spans entire array → one file per array.
+        # Must be a multiple of the inner chunk size.
+        shard_first = ((data.shape[0] + first_dim - 1) // first_dim) * first_dim
+        shards = (shard_first, *data.shape[1:])
+        arr = group.create_array(
+            path, shape=data.shape, dtype=data.dtype, chunks=chunks, shards=shards
+        )
+        arr[:] = data
+    elif _zarr.__version__.startswith("2"):
+        group.create_dataset(path, data=data, chunks=chunks)
+    else:
+        # zarr v3 library with zarr_format=2 — no sharding
+        arr = group.create_array(path, shape=data.shape, dtype=data.dtype, chunks=chunks)
+        arr[:] = data
+
+
 def write_id_arrays(
     geff_store: StoreLike,
     node_ids: np.ndarray,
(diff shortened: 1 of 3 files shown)


packages/geff/src/geff/core_io/_base_write.py
```diff
@@ -307,6 +307,74 @@ def write_arrays(
...
+    import zarr as _zarr
+
+    if data.shape[0] == 0:
+        group[path] = data
+        return
+
+    first_dim = _compute_first_dim_chunk(data.shape, data.dtype.itemsize)
+    chunks = (first_dim, *data.shape[1:])
+
+    use_sharding = (
```
cmalinmayor: We have a zarr utils function for detecting the zarr format version in core_io/utils.py - could use here, and it also works on any store not just the group

packages/geff/src/geff/core_io/_base_write.py
```diff
@@ -307,6 +307,74 @@ def write_arrays(
...
+    row_bytes = int(np.prod(shape[1:])) * itemsize if len(shape) > 1 else itemsize
+    n_rows = _TARGET_CHUNK_BYTES // max(row_bytes, 1)
+    # Round down to nearest power of two
+    if n_rows >= 1:
+        n_rows = 1 << (n_rows.bit_length() - 1)
+    return max(1, min(first_dim, n_rows))
+
+
+def _write_zarr_array(
+    group: Any,
```
cmalinmayor: What is thea ctual required type? Is it a general StoreLike, or does it need to be zarr.group?

cmalinmayor (changes requested):
Generally looks really good, thanks! I would love a test for the bug that would have failed on main and succeeded here. If I get the chance today I'll add that, otherwise if you beat me to it thats great as well.  Also, a minor comment about our zarr utilities/typing.

packages/geff/tests/test_core_io/test_base_read.py
```diff
@@ -226,6 +226,38 @@ def test_read_edge_props() -> None:
     _ = NxBackend.construct(**in_memory_geff)
 
 
+def test_build_w_masked_nodes_varlength() -> None:
```
JoOkuma, line 255: Thanks for the review @cmalinmayor . This is the regression test you mentioned above. Nonetheless, don't hesitate to add more tests.

I'll address your comments soon.

packages/geff/src/geff/core_io/_base_write.py
```diff
@@ -302,6 +302,74 @@ def write_arrays(
...
+    On zarr v3 sharding is used so every array becomes a single file: the shard
+    (outer chunk) spans the entire first dimension while inner sub-chunks are
+    ~8 MiB each.
+
+    Args:
+        group: The zarr group to write into.
+        path: The path within the group.
+        data: The numpy array to write.
+    """
+    import zarr as _zarr
```
cmalinmayor: Why re-importing this here? Can we just use the import at the top?

cmalinmayor:
Added a test just to make coverage happy since technically it was correct that we didn't test those two lines... 🙃 Otherwise you are correct the tests are great! Just missed it somehowd

packages/geff/src/geff/core_io/_base_write.py
JoOkuma: yes

packages/geff/src/geff/core_io/_base_write.py
JoOkuma: It's a zarr.Group, it doesn't work with `StoreLike` because it can be indexed by `path`.

JoOkuma:
Hi @cmalinmayor, I addressed your comments, lmk if I missed anything

packages/geff/src/geff/core_io/_base_write.py
```diff
@@ -302,6 +304,70 @@ def write_arrays(
...
+    row_bytes = int(np.prod(shape[1:])) * itemsize if len(shape) > 1 else itemsize
+    n_rows = _TARGET_CHUNK_BYTES // max(row_bytes, 1)
+    # Round down to nearest power of two
+    if n_rows >= 1:
+        n_rows = 1 << (n_rows.bit_length() - 1)
+    return max(1, min(first_dim, n_rows))
+
+
+def _write_zarr_array(
+    group: Any,
```
cmalinmayor:
```suggestion
    group: zarr.Group,
```

cmalinmayor (approved):
Thank you!! 🦖 🍅 🎉

PR #420, opened 2026-05-13
Switch to bash shell for running tests in CI to make windows tests fail

### LINKED ISSUE #332: Windows tests have a green check even when they fail
Brief google indicates something with the exit codes not being as expected

msschwartz21:
Closes #332

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
     name: Type checking
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
       - uses: astral-sh/setup-uv@v7
         with:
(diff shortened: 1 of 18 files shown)


cmalinmayor:
Do we want to make them pass here too? Or just the fix to make the CI fail when the tests fail?

msschwartz21:
> Do we want to make them pass here too? Or just the fix to make the CI fail when the tests fail?

I had wanted to fix the bug, but I guess it's not the end of the world to leave them failing and fix it in another PR. Do you have a preference?

cmalinmayor:
Looks like its a spatial-graph bug that needs to be fixed upstream - let's just mark spatial_graph on Windows as xfail for now.

cmalinmayor (approved):
Green check mark! Finally!!! 🍅 👍 🎉

PR #429, opened 2026-07-23
Add geffception to GeffMetadata through RelatedObjects

msschwartz21:
# Proposed Change
This PR adds metadata support for recording geffception. It does not provide any implementation support.

To do:

- [x] Write documentation about geffception
- [x] Decide about how to handle the prop attribute (see open questions)

# Open Questions
I'm tagging people who I think may have an opinion based on previous geffception conversations but please tag others if I forget anyone. @cmalinmayor @tinevez @DragaDoncila @lxenard 
- ~~Should we use `label_prop` to record the node prop that maps between the geffception geff and the core geff? Or should we create a new generic `prop` attribute?~~ Deprecated `label_prop` and replaced with `node_prop`. Let me know if we should also include an `edge_prop`, but I think we should specify if the prop applies to nodes or edges.
- ~~Should the prop attribute be required for geffception?~~ No
- Help with the documentation would be appreciated! I wrote something but it could be better 😅 

# Types of Changes
- New feature or enhancement
- Documentation update

- Specification
- Metadata

# Checklist

- [N/A] I have updated implementations to reflect the change. (This can happen in separate PRs on a feature branch, but must be complete before merging into main.)

### DIFF
--- a/docs/geffception.md
+++ b/docs/geffception.md
@@ -1,0 +1,129 @@
+# Geffception
+
+The GEFF specification supports hierarchical nesting of GEFF files. This is particularly useful in the context of cell tracking where some properties may be computed for an entire tracklet or lineage as opposed to for an individual node or edge.
+
+In the example illustrated below, `tracking_graph.geff` contains nodes with the properties `track_id`, `t`, `y`, and `x` and edges with the property `distance`. Within the folder structure of `tracking_graph.geff` is `lineage.geff`. The `GeffMetadata` of `tracking_graph.geff` points to `lineage.geff` as a related object whose node ids correspond to the `track_id` property of nodes in `tracking_graph.geff`.
+
+```python
+/path/to.zarr
+    tracking_graph.geff/
+        .zattrs # graph metadata with `geff_version` and RelatedObject metadata for lineage.geff
+        nodes/ 
+            ids # shape: (N,) dtype: uint64
+            props/
+                track_id/
+                    values # shape: (N,) dtype: uint16
+                t/
+                    values # shape: (N,) dtype: uint16
+                y/
+                    values # shape: (N,) dtype: float32
+                x/
+                    values # shape: (N,) dtype: float32
+        edges/
+            ids # shape: (E, 2) dtype: uint64
+            props/
+                distance/
+                    values # shape: (E,) dtype: float32
+        lineage.geff/
+            .zattrs # geff metadata
+            nodes/
+                ids # shape: (L,) dtype: uint64
+                props/
+                    track_displacement/
+                        values # shape: (L,) dtype: float32
+            edges/
+                ids # shape: (0,) dytpe: uint64
+```
+
+```jsonc
+// /path/to.zarr/tracking_graph.geff/.zattrs
+{
+    "geff": {
+        "directed": true,
+        "geff_version": "1.2.0.1.1",
+        "axes": [
+            { 
+                "name": "t", 
+                "type": "time",
+                "unit": "second", 
+                "min": 0, 
+                "max": 125 
+            },
+            {
+                "name": "x",
+                "type": "space",
+                "unit": "micrometer",
+                "min": 764.42,
+                "max": 2152.3
+            },
+            {
+                "name": "y",
+                "type": "space",
+                "unit": "micrometer",
+                "min": 81.667,
+                "max": 1877.7
+            }
+        ],
+        "node_props_metadata": {
+            "track_id": {
+                "identifier": "track_id",
+                "dtype": "uint16",
+                "varlength": false
+            },
+            "t": {
+                "identifier": "t",
+                "dtype": "uint16",
+                "varlength": false,
+                "unit": "second"
+            },
+            "y": {
+                "identifier": "y",
+                "dtype": "float32",
+                "varlength": false,
+                "unit": "micrometer"
+            },
+            "x": {
+                "identifier": "x",
+                "dtype": "float32",
+                "varlength": false,
+                "unit": "micrometer"
+            }
+        },
+        "edge_props_metadata": {
+            "distance": {
+                "identifier": "distance",
+                "dtype": "float32",
+                "varlength": false
+            }
+        },
+        "track_node_props": {
+            "lineage": "track_id"
+        },
+        "related_objects": [
+            {
+                "type": "geff",
+                "path": "lineage.geff",
+                "node_prop": "track_id"
+            }
+        ]
+    }
+}
+```
+
+```jsonc
+// /path/to.zarr/tracking_graph.geff/lineage.geff/.zattrs
+{
+    "geff": {
+        "directed": true,
+        "geff_version": "1.2.0.1.1",
+        "node_props_metadata": {
+            "track_displacement": {
+                "identifier": "track_displacement",
+                "dtype": "float32",
+                "varlength": false
+            }
+        },
+        "edge_props_metadata": {}
+    }
+}
+```
--- a/docs/geffception.md
+++ b/docs/geffception.md
@@ -1,6 +1,6 @@
-# Geffception
+# GEFFception: Hierarchical nesting of GEFFs
 
-The GEFF specification supports hierarchical nesting of GEFF files. This is particularly useful in the context of cell tracking where some properties may be computed for an entire tracklet or lineage as opposed to for an individual node or edge. For a complete example of a tracking graph that uses geffception to store tracklet and lineage properties, see [the tracking docs](./tracking.md).
+The GEFF specification supports hierarchical nesting of GEFF files. This is particularly useful in the context of cell tracking where some properties may be computed for an entire tracklet or lineage as opposed to for an individual node or edge. For a complete example of a tracking graph that uses GEFFception to store tracklet and lineage properties, see [the tracking docs](./tracking.md).
 
 In the example illustrated below, `tracking_graph.geff` contains nodes with the properties `track_id`, `t`, `y`, and `x` and edges with the property `distance`. Within the folder structure of `tracking_graph.geff` is `lineage.geff`. The `GeffMetadata` of `tracking_graph.geff` points to `lineage.geff` as a related object whose node ids correspond to the `track_id` property of nodes in `tracking_graph.geff`.
 
--- a/docs/geffception.md
+++ b/docs/geffception.md
@@ -1,6 +1,6 @@
 # Geffception
 
-The GEFF specification supports hierarchical nesting of GEFF files. This is particularly useful in the context of cell tracking where some properties may be computed for an entire tracklet or lineage as opposed to for an individual node or edge.
+The GEFF specification supports hierarchical nesting of GEFF files. This is particularly useful in the context of cell tracking where some properties may be computed for an entire tracklet or lineage as opposed to for an individual node or edge. For a complete example of a tracking graph that uses geffception to store tracklet and lineage properties, see [the tracking docs](./tracking.md).
 
 In the example illustrated below, `tracking_graph.geff` contains nodes with the properties `track_id`, `t`, `y`, and `x` and edges with the property `distance`. Within the folder structure of `tracking_graph.geff` is `lineage.geff`. The `GeffMetadata` of `tracking_graph.geff` points to `lineage.geff` as a related object whose node ids correspond to the `track_id` property of nodes in `tracking_graph.geff`.
 
--- a/docs/tracking.md
+++ b/docs/tracking.md
@@ -14,9 +14,9 @@ The `tracklet` and `lineage` properties specified in the `track_node_props` sect
 
 By providing and enforcing a definition of tracklet and lineage, and ensuring tracklet and lineage IDs can be exchanged rather than requiring them to be recomputed, we can ensure consistency of downstream analyses on the tracks across different tools.
 
-### Geffception for tracking
+### GEFFception for tracking
 
-In order to store properties that apply to an entire tracklet or lineage as opposed to an individual node or edge, we recommend [geffception](./geffception.md). Geffception allows for hierarchical nesting of geffs. 
+In order to store properties that apply to an entire tracklet or lineage as opposed to an individual node or edge, we recommend [geffception](./geffception.md). GEFFception allows for hierarchical nesting of geffs. 
 
 In the example illustrated below, there are three levels of nested geffs.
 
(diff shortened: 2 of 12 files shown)


cmalinmayor:
I think we should re-use label_prop. We can do a backward compatible rename where both `prop` and `label_prop` are valid, but `label_prop` is deprecated, or just accept that the name makes less sense for graph-type related objects.

msschwartz21:
@tlambert03 Would you mind taking a quick look at how I deprecated `label_prop` and see if I did it "correctly" from a pydantic perspective?

lxenard:
Geffception! 🎉
I agree with the changes, and I will check the documentation more thoroughly tomorrow.

One question/remark: in TrackMate and Pycellin's cases, we have both tracklet props and lineage props, meaning we need to do a double geffception to store all the data. In my demo, I split the `geff` field of `related_object` into `tracklet_geff` and `lineage_geff`. It's not perfect because it constrains the type of geff in a geffception. On the other hand, what you did allow users to store any type of geff. I can still store several geffs by putting them inside one another: regular geff > tracklet geff > lineage geff, which has the added benefit of keeping a pyramidal modeling of the data. So I really think your version is better even if even more 🤯. I'm just trying to think of other uses for geffception to see if it could cause issues down the line, but my mind is drawing a blank. Any thoughts?

lxenard:
And what do you mean by: "Should the prop attribute be required for geffception?"

msschwartz21:
> And what do you mean by: "Should the prop attribute be required for geffception?"

Sorry I should have updated prop to node_prop. In the RelatedObjects validation, we could require that if type=='geff' that a node_prop is specified such that a geff RelatedObject can always be mapped back to a node property on the core geff. I actually don't think this is a necessary restriction.

> One question/remark: in TrackMate and Pycellin's cases, we have both tracklet props and lineage props, meaning we need to do a double geffception to store all the data. In my demo, I split the geff field of related_object into tracklet_geff and lineage_geff. It's not perfect because it constrains the type of geff in a geffception. On the other hand, what you did allow users to store any type of geff. I can still store several geffs by putting them inside one another: regular geff > tracklet geff > lineage geff, which has the added benefit of keeping a pyramidal modeling of the data. So I really think your version is better even if even more 🤯. 

Alternatively when you need to have both tracklet and lineage properties could you keep them on the same level, e.g. 
```
main.geff/
  nodes
  edges
  tracklet.geff/
  lineage.geff/
```
And then the metadata of main.geff would define `RelatedObject(type='geff', path='tracklet.geff', node_prop='tracklet_id')` and `RelatedObject(type='geff', path='lineage.geff', node_prop='lineage_id')`. 
  
> I'm just trying to think of other uses for geffception to see if it could cause issues down the line, but my mind is drawing a blank. Any thoughts?

I have been going through the same exercise and haven't come up with specific examples that would break in the current framework. However, I generally decided to err on the side of less restrictive with the hope that it's less likely to cause problems in the future.

tinevez:
> Geffception! 🎉 I agree with the changes, and I will check the documentation more thoroughly tomorrow.
> 
> One question/remark: in TrackMate and Pycellin's cases, we have both tracklet props and lineage props, meaning we need to do a double geffception to store all the data. In my demo, I split the `geff` field of `related_object` into `tracklet_geff` and `lineage_geff`. It's not perfect because it constrains the type of geff in a geffception. On the other hand, what you did allow users to store any type of geff. I can still store several geffs by putting them inside one another: regular geff > tracklet geff > lineage geff, which has the added benefit of keeping a pyramidal modeling of the data. 

> Alternatively when you need to have both tracklet and lineage properties could you keep them on the same level, e.g.
> 
> main.geff/
>   nodes
>   edges
>   tracklet.geff/
>   lineage.geff/


Well since we have the Geffception, we might as well go all the way. So I would humbly favor the nested structure. 
 I really like the nesting 
```lineage_geff ⊂ tracklet_geff ⊂ core_geff ```

Here is an example. We could have something like:

CORE_GEFF
12 nodes, 11 edges

tracklet_id == the id of nodes in the TRACKLET_GRAPH

```
    a1(tracklet_id=1)
    |
    a2(tracklet_id=1)
    |
    +--------------------------+
    |                          |
    b1(tracklet_id=2)          c1(tracklet_id=3)
    |                          |
    b2(tracklet_id=2)          c2(tracklet_id=3)
                               +--------------------------+
                               |                          |
                               d1(tracklet_id=4)          e1(tracklet_id=5)
                               |                          |
                               d2(tracklet_id=4)          e2(tracklet_id=5)
                               |                          |
                               d3(tracklet_id=4)          e3(tracklet_id=5)
```

TRACKLET_GEFF
5 nodes, 4 edges

lineage_id == the id of nodes in the LINEAGE_GEFF

```
a(id=1, lineage_id=1)
|
+----------------------+
|                      |
b(id=2, lineage_id=1)  c(id=3, lineage_id=1)
                       +----------------------+
                       |                      |
                       d(id=4, lineage_id=1)  e(id=5, lineage_id=1)
```

LINEAGE_GEFF
1 node, 0 edges

```
track_1(id=1)
```
(please admire my hand-crafted ascii binary tree)

With this, the core GEFF can still access lineage properties by recursing in the tracklet GEGG. And the tracklet GEFF can be loaded and studied separately, and still make sense of the lineage properties. 

If the tracklet GEFF is missing and the core GEFF only contains the lineage GEFF, it can still see the lineage properties via the lineage related obj.

**Comment by lxenard** [comment]
> Sorry I should have updated prop to node_prop. In the RelatedObjects validation, we could require that if type=='geff' that a node_prop is specified such that a geff RelatedObject can always be mapped back to a node property on the core geff. I actually don't think this is a necessary restriction.

I agree. Even if right now I don't see a use for a geffception not mapped to the main geff, better to err on the side of caution and not be too restrictive. It's easy enough to check for the mapping.
But maybe, in term of validation, if a `node_prop` is set in the the `geff` RelatedObject, we could enforce that the prop actually exist in the main geff?

> And then the metadata of main.geff would define `RelatedObject(type='geff', path='tracklet.geff', node_prop='tracklet_id')` and `RelatedObject(type='geff', path='lineage.geff', node_prop='lineage_id')`.

Ah my bad, I don't know why but I though we couldn't have 2 RelatedObject with the same type 🤷

**Comment by lxenard** [comment]
> (please admire my hand-crafted ascii binary tree)

Seems like I still have a lot to learn from you! 😂

> With this, the core GEFF can still access lineage properties by recursing in the tracklet GEGG. And the tracklet GEFF can be loaded and studied separately, and still make sense of the lineage properties.

I've been debating with myself since yesterday evening on wether doing flat or multi-nested geffception, but this actually shifts the balance towards "multi-nested is better". In case a user is not interested in the main geff, they can read the tracklet geff folder to get both tracklet and lineage props, without needing to specifically check for another geff at the same directory level. Even from a reader/writer perspective, I feel like it makes more sense.

File: docs/tracking.md [hunk 1]
```diff
@@ -14,4 +14,6 @@ The `tracklet` and `lineage` properties specified in the `track_node_props` sect
 
 By providing and enforcing a definition of tracklet and lineage, and ensuring tracklet and lineage IDs can be exchanged rather than requiring them to be recomputed, we can ensure consistency of downstream analyses on the tracks across different tools.
 
+In order to store properties that apply to an entire tracklet or lineage as opposed to an individual node or edge, please see [geffception](./geffception.md)
```
lxenard:
```suggestion
In order to store properties that apply to an entire tracklet or lineage as opposed to an individual node or edge, please see [geffception](./geffception.md).
```

geff-schema.json
```diff
@@ -509,8 +509,22 @@
...
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Property name that links between the related object and node properties. For segmentation data, this field should refer to the node property that specifies the segmentation ID for each node. For geffception, this property typically links between the node ID of the geffception graphs and the property that specifies which geffception graph a node in the core graph belongs to.",
```
lxenard: Not easy to understand but I don't have a better proposition without making it too long. In any case, users interested in geffception will either already know what it is about and just need a reminder on how to use this field, or if they're not familiar they will lookup the full documentation. So I think it's good as is. Or maybe add a:
"See documentation for a full (concrete?) example."

packages/geff-spec/src/geff_spec/_schema.py
```diff
@@ -98,17 +98,40 @@ class RelatedObject(BaseModel):
...
+        deprecated="Deprecated in geff-spec v1.2.1 in favor of `node_prop`.",
+    )
+    node_prop: str | None = Field(
+        default=None,
+        description=(
+            "Property name that links between the related object and node properties. "
+            "For segmentation data, this field should refer to the node property that "
+            "specifies the segmentation ID for each node. For geffception, this property "
+            "typically links between the node ID of the geffception graphs and the property "
+            "that specifies which geffception graph a node in the core graph belongs to."
```
lxenard: See my comment in geff-schema.json

lxenard (commented):
All good for me! The geffception documentation was clear and to the point.
Depending on if you decide on single-nested or multi-nested geffs, we may have to add a more complex example.

msschwartz21:
Thanks for the feedback and discussion! Those ascii trees make a compelling argument for multi-nested approach 😂 

I think I'll add a section to to the tracking documentation page explaining the multi-nested geff convention when both lineages and tracklets are included. In my opinion this is a "should" type thing, not a "must" thing.

I am going to shamelessly steal those trees for the docs 😅

msschwartz21:
@lxenard would you mind making one last pass through the docs?

docs/tracking.md
```diff
@@ -14,6 +14,79 @@ The `tracklet` and `lineage` properties specified in the `track_node_props` sect
...
+```
+
+`core.geff` contains 12 nodes and 11 edges. Nodes have the property `tracklet_id` which corresponds to the ids of nodes in `tracklet.geff`. `core.geff` contains the following metadata and graph structure shown below:
+
+```jsonc
+"related_objects": [
+    {
+        "type": "geff",
+        "path": "tracklet.geff",
+        "node_prop": "track_id"
```
lxenard:
```suggestion
        "node_prop": "tracklet_id"
```

tlambert03:
> @tlambert03 Would you mind taking a quick look at how I deprecated `label_prop` and see if I did it "correctly" from a pydantic perspective?

I'm so sorry for delay!  Busy days here.  If this is still open tomorrow; I'll have a look!

lxenard:
@msschwartz21 Done! Thank you for this 🙇‍♀️
I will update Pycellin accordingly once I have submitted my thesis ah ah ah

DragaDoncila:
Apologies for the drive-by, and I will do my best to come back to this in more detail. Generally speaking I'm very happy with this PR. I'm a little confused by the `node_prop` choice. I understand previously, `label_prop` was the name of the property that held the segmentation labels, but now that its `node_prop` I feel that it's a bit ambiguous. 

@msschwartz21 what are the semantics of the `node_prop`?

packages/geff-spec/src/geff_spec/_schema.py
```diff
@@ -235,14 +266,16 @@ class GeffMetadata(BaseModel):
...
+            "be used for label and image objects, respectively. 'geff' should be used to "
+            "refer to geffception. See geffception docs for more details. Other types are "
+            "also allowed. "
             "The 'path' should be relative to the geff zarr-attributes file. "
             "It is strongly recommended all related objects are stored as siblings "
             "of the geff group within the top-level zarr group. "
-            "The 'label_prop' is only valid for type 'labels' and specifies the node property "
-            "that will be used to identify the labels in the related object. "
+            "The deprecated 'label_prop' is only valid for type 'labels' and specifies the node"
+            "property that will be used to identify the labels in the related object. "
```
DragaDoncila:
```suggestion
            "property that will be used to identify the nodes in the related object. "
```

geff-schema.json
```diff
@@ -509,8 +509,22 @@
...
+          "anyOf": [
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Property name that links between the related object and node properties. For segmentation data, this field should refer to the node property that specifies the segmentation ID for each node. For geffception, this property typically links between the node ID of the geffception graphs and the property that specifies which geffception graph a node in the core graph belongs to. See the geffception page in the documentation for a concrete example.",
```
DragaDoncila:
```suggestion
          "description": "Property on current geff graph nodes that link to the related object. For segmentation data, this field should refer to the node property that specifies the segmentation ID for each node. For geffception, this property typically links a node property on the current graph to node IDs of the related, geffception graphs. See the geffception page in the documentation for a concrete example.",
```

DragaDoncila (commented):
Ok after reading a bit more I do get it now. I've suggested one correction, and one rewording -- for me the reworded description is a bit easier to understand but I'm not married to it if other folks prefer the current wording!

tlambert03:
I took a look and it all seems good to me.  There's more than one way, and I wouldn't bike shed it.  This works fine

docs/geffception.md
```diff
@@ -0,0 +1,129 @@
+# Geffception
```
cmalinmayor:
```suggestion
# Geffception: Hierarchical nesting of GEFFs
```

docs/tracking.md
```diff
@@ -14,4 +14,79 @@ The `tracklet` and `lineage` properties specified in the `track_node_props` sect
 
 By providing and enforcing a definition of tracklet and lineage, and ensuring tracklet and lineage IDs can be exchanged rather than requiring them to be recomputed, we can ensure consistency of downstream analyses on the tracks across different tools.
 
+### Geffception for tracking
```
cmalinmayor: GEFFception? 🤔

docs/tracking.md
```diff
@@ -14,4 +14,79 @@ The `tracklet` and `lineage` properties specified in the `track_node_props` sect
...
+```
+core.geff/
+    tracklet.geff/
+        lineage.geff/
+```
+
+`core.geff` contains 12 nodes and 11 edges. Nodes have the property `tracklet_id` which corresponds to the ids of nodes in `tracklet.geff`. `core.geff` contains the following metadata and graph structure shown below:
+
+```jsonc
+"related_objects": [
```
cmalinmayor, line 32: Doesnt this path mean next to the root? Or inside? I'm confused

cmalinmayor (changes requested):
Other than my confusion about if related paths are next to the geff or inside it by default, this looks good!

msschwartz21:
> Other than my confusion about if related paths are next to the geff or inside it by default, this looks good!

I was also confused by this and I will try to figure out if we defined that explicitly. If not we should.

#430

docs/tracking.md
msschwartz21: Yeah I noticed we're not really consistent on GEFF vs `geff`. Might be worth a round of edits in the future. But I think I agree with you for this case.

cmalinmayor (approved):
🍅

